PR #4, opened 2017-05-10
[DROOLS-1521] Add DSL script for downstream PR jobs

psiroky:
@mbiarnes please take a look whenever you have some time

### DIFF
--- a/job-dsls/jobs/downstream_pr_jobs.groovy
+++ b/job-dsls/jobs/downstream_pr_jobs.groovy
@@ -1,0 +1,190 @@
+/**
+ * Creates downstream pullrequest (PR) jobs for appformer (formerly known as uberfire), dashbuilder and kiegroup GitHub org. units.
+ * These jobs execute the full downstream build for a specific PR to make sure the changes do not break the downstream repos.
+ */
+import org.kie.jenkins.jobdsl.Constants
+
+def final DEFAULTS = [
+        ghOrgUnit              : "kiegroup",
+        branch                 : "master",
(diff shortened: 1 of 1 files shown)


mbiarnes:
@psiroky why are droolsjbpm-tools, kie-docs and kie-wb-distributions excluded?

psiroky:
@mbiarnes because no other repos depend on them, so there is no downstream build required. I thought the comment explains that clearly, but maybe it could be enhanced.

mbiarnes:
@psiroky well explained - it is more that I didn't understand this well, my personal groovy lack- I understood that these are the job to creates a build for each repostory when a PR is raised - so in my understanding (I repeat in my understanding and this could be wrong) I see that droolsjbpm-tools, kie-docs and kie-wb-distributions are excluded - so there is no PR-job for them? But for sure I don't see where all <reps> are defined   - I thought they are listed in REPO_CONFIGS

psiroky:
@mbiarnes we already have "standard" PR jobs for all the repos. See https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/jobs/pr_jobs.groovy. These changes are a new set of jobs, which run the whole downstream build (not just the build for the repo for which the PR was created). For some repos it does not make sense to execute the downstream build, so those are excluded (actually commented right now, but we can remove them completely as well).

psiroky:
@mbiarnes could you please check this again? Is there anything else I should update before merging?

mbiarnes:
@psiroky looks good +1

PR #1, opened 2017-05-10
added new script for kie releases

mbiarnes:

PR #6, opened 2017-05-11
added missing plugins for tests

mbiarnes:


mbiarnes:
@psiroky added missing plugins so PR #1 will build

PR #10, opened 2017-05-18
Archive only the required WARs

psiroky:
We only need to archive the actual end-result WARs. The old
GLOB matches also e.g. kie-wb-webapp.war or kie-wb-common-showcase.war
which are not needed and only take up storage space.


psiroky:
@mbiarnes could you please review this?

PR #13, opened 2017-06-22
Make sure the Selenium tests are executed for PR jobs

psiroky:
The app (WAR) needs to be specified using the profile, otherwise
the tests are skipped. This is a recent configuration change, which
was forgotten to be reflected in the job configs.


psiroky:
CC @jhrcek

PR #17, opened 2017-07-19
added new script and adapt kie_releases_jobs.groovy to use the right …

mbiarnes:
…branches for master
@psiroky when someone merges this new script we should disable the Trigger kieAll for master to prevent that we have the execution twice. 
I would suggest we wait with the disabling until tomorrow so we can see if the new script works.

### DIFF
--- a/job-dsls/build.gradle
+++ b/job-dsls/build.gradle
@@ -48,6 +48,7 @@ dependencies {
 
     // plugins to install in test instance (all plugins used by the DSL scripts needs to be specified)
     // keep the alphabetical order
+    testPlugins 'com.cloudbees.plugins:build-flow-plugin:0.20'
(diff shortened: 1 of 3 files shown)


mbiarnes:
@psiroky can you approve and merge it?

psiroky:
@mbiarnes how urgent is it? There seems some duplication we could get rid of, but that would take some time to review. If its needed now, I will just merge it as is.

mbiarnes:
@psiroky better to have everything working by DSL scripts then by parts - I want to replace the kieAll build master with this. I know well that there are a bunch of things we can get rid of in future

psiroky:
Yeah, I am just afraid that once it gets merged it will never get fixed...but let's hope it will.

PR #24, opened 2017-07-26
force to exit script if commnad was not successful

mbiarnes:


job-dsls/jobs/FlowJob.groovy
```diff
@@ -393,6 +393,12 @@ DEPLOY_DIR=$WORKSPACE/deploy-dir
 ./droolsjbpm-build-bootstrap/script/mvn-all.sh -B -e clean deploy -T1C -Dfull -Drelease -DaltDeploymentRepository=local::default::file://$DEPLOY_DIR -s $SETTINGS_XML_FILE\\
  -Dkie.maven.settings.custom=$SETTINGS_XML_FILE -Dmaven.test.redirectTestOutputToFile=true -Dmaven.test.failure.ignore=true -Dgwt.compiler.localWorkers=1\\
  -Dgwt.memory.settings="-Xmx4g -Xms1g -Xss1M"
+#force to exit the script if the provious command fails
+scriptStatus=$?
+if [ $scriptStatus -ne 0 ]; then
+   echo "build failed for some reason"
+   exit
+fi
```
psiroky: This is not really elegant. I think it is better to use `set -e`.

PR #28, opened 2017-07-31
Archive kie-server WARs in downstream PR builds

psiroky:
@mbiares please take a look.

CC @hasys


psiroky:
There is no WildFly specific WAR. The `ee7` is the one to use.

psiroky:
I will give this a test ride on master and then back port to 7.2.x and 6.5.x

PR #37, opened 2017-08-07
Archive additional target/*.log files

psiroky:
@mbiarnes please take a look

/CC @desmax74  - this should make sure the log files are always archived as a build artifacts.


desmax74:
+1 @psiroky "**/target/*.log" cover our needs

PR #40, opened 2017-08-08
upgraded MAVEN version for additional tests

mbiarnes:


job-dsls/jobs/kie_release_jobs.groovy
```diff
@@ -6,6 +6,7 @@ def dashbuilderVersion="0.9.x"
 def javadk="jdk1.8"
 def jaydekay="JDK1_8"
 def mvn="APACHE_MAVEN_3_3_9"
+def mvnVersionTest="apache-maven-3.3.9"
```
psiroky: Why `Test` in the var name?

mbiarnes: because this is version apache-maven-3.3.9 is only required in the additional tests - if you don't like this variable and I should name it otherwise please tell me

mbiarnes:
due to conflicts that I couldn't resolve (it didn't work git rebase --continue after correcting conflicts) and because of the loss of time to get this sorted out I prefer to do a new PR

PR #43, opened 2017-08-10
Upgrade maven master

mbiarnes:
removed also leftover in job-dsls/jobs/kie_release_jobs.groovy - def pushReleaseBranches=""" ... """" since this is in scripts droolsjbpm-build-bootstrap/script/release/kie-createAndPushReleaseBranches.sh

### DIFF
--- a/job-dsls/jobs/kie_release_jobs.groovy
+++ b/job-dsls/jobs/kie_release_jobs.groovy
@@ -5,8 +5,9 @@ def uberfireVersion="2.0.x"
 def dashbuilderVersion="1.0.x"
 def javadk="jdk1.8"
 def jaydekay="JDK1_8"
-def mvn="APACHE_MAVEN_3_3_9"
-def mvnHome="${mvn}_HOME"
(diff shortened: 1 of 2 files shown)


job-dsls/jobs/kie_release_jobs.groovy
```diff
@@ -458,7 +389,7 @@ matrixJob("serverMatrix-kieReleases-${kieVersion}") {
             properties("maven.test.failure.ignore": true)
             properties("deployment.timeout.millis":"240000")
             properties("container.startstop.timeout.millis":"240000")
-            properties("eap64x.download.url":"http://download.devel.redhat.com/released/JBEAP-6/6.4.4/jboss-eap-6.4.4-full-build.zip")
+            properties("eap7.download.url":"http://download.eng.brq.redhat.com/released/JBEAP-7/7.0.2/jboss-eap-7.0.2-full-build.zip")
```
psiroky: Let's use the generic download.devel. The specific server will be picked up based on the location. Downloading stuff from BRQ into RDU will be slower.

PR #53, opened 2017-08-16
disabled all jobs for 7.2.x

mbiarnes:


job-dsls/jobs/FlowJob.groovy
```diff
@@ -140,6 +140,8 @@ job("ErraiFor_kieAllBuild_${kieMainBranch}") {
         stringParam("erraiBranch", "errai branch", "Branch of errai. This will be usually set automatically by the parent trigger job. ")
     }
 
+    disabled(shouldDisable = true)
```
psiroky: Simply `disabled()` would have sufficed, but that's just cosmetics.

PR #57, opened 2017-08-21
Use only two threads when building kie-wb-distributions

psiroky:
kie-wb-distributions will soon contain third GWT webapp, which means
the build will start to fail with OOM. This is a quick fix which
will make sure only two parallel threads are used to build the
kie-wb-distributions repo.

CC @cristianonicolai


cristianonicolai:
+1, thanks @psiroky

cristianonicolai:
@mbiarnes are you ok with these changes?

Issue #64, opened 2017-09-04
Replace Maven jobs with Freestyle jobs


psiroky:
The standard 'deploy' jobs are still using the Maven job. It adds a lot of hooks into the build and sometimes even changes the way a build is being executed. This sometimes leads to an unexpected failures, where the build works fine with just "mvn clean install", but fails in the job. These jobs need to be replaced with the Freestyle jobs, which do the Maven build exactly as expected. We can not simply update the config, Jenkins won't accept that. The old jobs first need to be deleted and new ones then re-created. We will loose the build history, but that's not a big issue.

List of issues with the Maven job type:
 - it partially ignores the `MAVEN_OPTS` settings. When parsing the POMs is simply does it as part of the `slave.jar` process. This results in OOME for big projects.

psiroky:
Fixed on master. PRs for 7.3.x and 6.5.x prepared and will be merged later today.

PR #66, opened 2017-09-04
[#64] use Freestyle job type for deploy jobs

psiroky:


psiroky:
I will merge this later today, probably around midnight CEST to minimize the disruptions.

PR #76, opened 2017-09-14
added jobs for automatic creation of jBPM-installers

mbiarnes:


mbiarnes:
only makes sense with: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/551

job-dsls/jobs/kie_release_jobs.groovy
```diff
@@ -737,11 +737,12 @@ job("copyBinariesToFilemgmt-kieReleases-${kieVersion}") {
 
     steps {
         environmentVariables {
-            envs(MAVEN_OPTS : "${mvnOpts}", MAVEN_HOME : "\$${mvnHome}", MAVEN_REPO_LOCAL : "/home/jenkins/.m2/repository", PATH : "\$${mvnHome}/bin:\$PATH")
+            envs(MAVEN_OPTS : "${mvnOpts}", MAVEN_HOME : "\$${mvnHome}", MAVEN_REPO_LOCAL : "/home/jenkins/.m2/repository", PATH : "\$WORKSPACE/droolsjbpm-build-bootstrap/script/release:\$${mvnHome}/bin:\$PATH")
```
psiroky: This looks suspicious. Why exactly is it needed?

mbiarnes: A script X calls another script Y. I didn't found another way to add this PATH variable for all scripts being present. If I don't use this variable scripts X ececutes but script Y is not found. Do you know a better way?

psiroky: I think we should either specify an absolute path or make sure the caller (X) can figure out precisely where the Y is located. Which script is this?

PR #77, opened 2017-09-15
added kie-soup to dailyBuilds jobs

mbiarnes:
@psiroky makes only sense if https://github.com/kiegroup/kie-soup/pull/2 was merged

### DIFF
--- a/job-dsls/jobs/FlowJob.groovy
+++ b/job-dsls/jobs/FlowJob.groovy
@@ -8,6 +8,7 @@ def mvnHome="${mvnToolEnv}_HOME"
 def mvnOpts="-Xms1g -Xmx3g"
 def kieMainBranch="master"
 def erraiBranch="master"
+def kiesoupBranch="master"
 def uberfireBranch="master"
 def dashbuilderBranch="master"
 def erraiVersionOld="4.1.0-SNAPSHOT"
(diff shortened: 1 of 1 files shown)


job-dsls/jobs/FlowJob.groovy
```diff
@@ -8,6 +8,7 @@ def mvnHome="${mvnToolEnv}_HOME"
 def mvnOpts="-Xms1g -Xmx3g"
 def kieMainBranch="master"
 def erraiBranch="master"
+def kiesoupBranch="master"
```
psiroky: I would suggest to use `$kieMainBranch`. Less maintenance for us.

job-dsls/jobs/FlowJob.groovy
```diff
@@ -192,6 +196,88 @@ job("errai-kieAllBuild-${kieMainBranch}") {
...
+# build the repos & deploy into local dir (will be later copied into staging repo)
+deployDir=$WORKSPACE/deploy-dir
+# (1) do a full build, but deploy only into local dir
+# we will deploy into remote staging repo only once the whole build passed (to save time and bandwith)
+mvn -U -B -e clean deploy -T2 -Dfull -Drelease -DaltDeploymentRepository=local::default::file://$deployDir -s $SETTINGS_XML_FILE\\
+ -Dmaven.test.failure.ignore=true -Dgwt.compiler.localWorkers=3
+# (2) upload the content to remote staging repo
+cd $deployDir
+mvn -B -e org.sonatype.plugins:nexus-staging-maven-plugin:1.6.5:deploy-staged-repository -DnexusUrl=https://repository.jboss.org/nexus -DserverId=jboss-releases-repository -DrepositoryDirectory=$deployDir\\
+ -s $SETTINGS_XML_FILE -DstagingProfileId=15c3321d12936e -DstagingDescription="kie-soup $kiesoupVersion" -DstagingProgressTimeoutMinutes=30'''
```
psiroky: We should somehow make sure that we can have these scripts outside of the DSL code. This makes them untestable on its own.

mbiarnes: @psiroky I have on my to do list to refactor all this

psiroky: You can not expect to have an explicit time for refactoring. That basically never happens. Refactoring needs to be continuously practiced in small bursts, and of course accounted in the estimates. Anyway, I will merge the PR since that's probably the only choice at this point.

PR #79, opened 2017-09-20
added script for creating *.json file

mbiarnes:


mbiarnes:
@psiroky the creation of json file is required from prod guys
@AntStephenson when this PR is merged the nightly builds will have a json file

AntStephenson:
@mbiarnes thanks very much for working on this, I look forward to integrating it! do you have some sample output from the latest version of the script?

job-dsls/jobs/FlowJob.groovy
```diff
@@ -405,8 +409,36 @@ mvn -B -e org.sonatype.plugins:nexus-staging-maven-plugin:1.6.5:deploy-staged-re
...
+   "repos" : [
+      {
+         $commitHash
+      }
+    ],
+   "source_product_tag":"$sourceProductTag",
+   "target_product_build":"$targetProductBuild" 
+   }
+}
+EOF
```
psiroky: Just thinking out loud: what about having this as a template file and then let the scripts fill in the variable content like the dates, tags, etc. WDYT?

job-dsls/jobs/FlowJob.groovy
```diff
@@ -405,8 +409,36 @@ mvn -B -e org.sonatype.plugins:nexus-staging-maven-plugin:1.6.5:deploy-staged-re
...
+# indent json
+python -m json.tool int.json >> $fileToWrite
```
psiroky: Following is mostly my personal opinion, so feel free to ignore it. I am not sure this indenting justifies the need to install and maintain Python on the slaves. I agree that we need some way to script this, but I was more thinking of using groovy.

In any case, we will need to make sure that either groovy or python are installed on the machines which will run this script.

AntStephenson: @psiroky - some useful comments! are there any cases where Python would not be on the slaves? afaik, there are a lot of built in system tools that are written in Python and I was always under the impression for this reason, it would always be available. IIRC, dnf for a long time was mostly Python.

Either way, I guess this or Groovy is fine, but being conscious of reducing maintenance, I think using built in Python libraries are probably OK. 

WDYT?

psiroky: TBH, I haven't checked, so it may very well be already there (coincidentally with something else). So if that's the case, then this is probably not an issue.

I do not like Python at all, but I'll not let this hold this PR and the other possible uses :)

psiroky: My apologies. The slaves do have `Python 2.7.5` installed as of now.

mbiarnes:
@psiroky is there something missing right now to merge this PR?

psiroky:
@mbiarnes I can't say I like the way the JSON is being created. However, that be improved later if deemed necessary.

PR #85, opened 2017-10-11
added job for automatic creation of jBPM-installers

mbiarnes:


mbiarnes:
@psiroky only makes sense with PR https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/574

job-dsls/jobs/kie_release_jobs.groovy
```diff
@@ -1148,6 +1213,7 @@ nestedView("kieReleases-${kieMainBranch}"){
                 name("removeReleaseBranches-kieReleases-${kieMainBranch}")
                 name("updateToNextDevelopmentVersion-kieReleases-${kieMainBranch}")
                 name("copyBinariesToFilemgmt-kieReleases-${kieMainBranch}")
+                name("jBPM-installer-kieReleases-${kieMainBranch}")
```
psiroky: Just thinking out loud: do we need a new job for this? Does it take too long? Or what was the reasoning behind creating a new job instead of updating the `copyBinariesToFilemgmt` one?

mbiarnes: @psiroky I will remove the job and add a call to kie-jbpmInstaller.sh in copyBinariesToFilemgmt

PR #90, opened 2017-10-25
Twaeks master

mbiarnes:


job-dsls/jobs/kie_release_jobs.groovy
```diff
@@ -722,7 +722,7 @@ job("copyBinariesToFilemgmt-kieReleases-${kieVersion}") {
 
     wrappers {
         timeout {
-            absolute(120)
+            absolute(180)
```
psiroky: I guess this is OK as a workaround, but if 2 hours is not enough to upload the binaries, then we need a proper fix. Is this related to a slow link to the jboss.org servers?

mbiarnes: @psiroky  I guess so - this takes longer then I assumed. it is simple a scp from kie-release to Nexus - it is to slow

PR #98, opened 2017-11-09
added jbpm-work-items to scripts

mbiarnes:
@mswiderski @tsurdilo the repo jbpm-work-items was added to the scripts

### DIFF
--- a/job-dsls/jobs/deploy_jobs.groovy
+++ b/job-dsls/jobs/deploy_jobs.groovy
@@ -143,6 +143,11 @@ def final REPO_CONFIGS = [
                 artifactsToArchive     : DEFAULTS["artifactsToArchive"] + [
                         "kie-wb-tests/kie-wb-tests-gui/target/screenshots/**"
                 ],
+                downstreamRepos        : ["jbpm-work-items"]
+        ],
+        "jbpm-work-items"                  : [
+                label                  : "rhel7 && mem4g",
+                ircNotificationChannels: ["#jbpmdev"],
                 downstreamRepos        : []
         ]
 ]
--- a/job-dsls/jobs/downstream_pr_jobs.groovy
+++ b/job-dsls/jobs/downstream_pr_jobs.groovy
@@ -62,7 +62,8 @@ def final REPO_CONFIGS = [
         "jbpm-designer"             : [],
         "jbpm-wb"                   : [],
         //"kie-docs"                  : [], // no other repo depends on kie-docs
-        //"kie-wb-distributions"      : [] // kie-wb-distributions is the last repo in the chain
+        "kie-wb-distributions"      : [],
+        "jbpm-work-items"           : [] // jbpm-work-items is the last repo in the chain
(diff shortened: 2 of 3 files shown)


job-dsls/jobs/downstream_pr_jobs.groovy
```diff
@@ -62,7 +62,8 @@ def final REPO_CONFIGS = [
         "jbpm-designer"             : [],
         "jbpm-wb"                   : [],
         //"kie-docs"                  : [], // no other repo depends on kie-docs
-        //"kie-wb-distributions"      : [] // kie-wb-distributions is the last repo in the chain
+        "kie-wb-distributions"      : [],
+        "jbpm-work-items"           : [] // jbpm-work-items is the last repo in the chain
```
psiroky: Since the `jbpm-work-items` is _not_ part of the `repository-list.txt` (and not if/when that would happen). As such we should include it in the downstream builds. I would revert this change completely - IMO we don't need downstream PR job for `kie-wb-distributions`, nor `jbpm-work-items`.

job-dsls/jobs/deploy_jobs.groovy
```diff
@@ -143,6 +143,11 @@ def final REPO_CONFIGS = [
                 artifactsToArchive     : DEFAULTS["artifactsToArchive"] + [
                         "kie-wb-tests/kie-wb-tests-gui/target/screenshots/**"
                 ],
+                downstreamRepos        : ["jbpm-work-items"]
+        ],
+        "jbpm-work-items"                  : [
+                label                  : "rhel7 && mem4g",
+                ircNotificationChannels: ["#jbpmdev"],
```
psiroky: I am afraid this is going to be a bit harder. The current jobs are building the needed upstream repos from sources, based on `branch-mapping.yaml` and `repository-list.txt`. The `jbpm-work-items` repo is not in any of those files, so the build will fail. I see two options here:

 1) rely on remote SNAPSHOTs (and don't build any upstream repos)
 2) "hack" the config and build the upstream repos up until jbpm (that should be enough)

I would probably vote for 2. as that will make the jobs a bit more stable (avoiding the remote SNAPSHOT dependency).

mbiarnes: @psiroky I don't know how to do this - I thought I am doing it right, but I see it was not.
Tbh I don't know what you mean with 2.
To prevent our endless loop of modifications until I do it it right I ask you please to do this.

job-dsls/jobs/downstream_pr_jobs.groovy
mbiarnes: @psiroky I think it is better to close the whole PR and do a new one. But as I commented before - I don't know how to do this well.

psiroky:
OK, looking at this now.

psiroky:
Replaced by https://github.com/kiegroup/kie-jenkins-scripts/pull/99

PR #100, opened 2017-11-10
Initial import of Ansible based OpenStack image creation for slaves

psiroky:


mbiarnes:
@psiroky is it important for kie-releases, kieci-01 and kieci-02 to have installed  packer and ansible?

psiroky:
@mbiarnes no. We won't be creating the images on those machines (hopefully). We can schedule a call or something and I can walk you through the contents of this PR.

mbiarnes:
@psiroky yes - this would be great
Meanwhile - I have to trust you blind as don't not know exactly how this will work. But if you want to merge this I can approve this.

AntStephenson:
@psiroky should we merge this now?

psiroky:
@AntStephenson would be great to get a proper review for these changes. Let's wait for our mtg and after that perhaps you or @pszubiak could take a look?

AntStephenson:
@psiroky  I've already looked through today, it looks great! I agree we should modulate some of the tasks to make it easier to consume. But we can do this later, I think. We can wait until after the meeting of course.

jenkins-slaves/ansible/data/config/pulp/pulp.repo
```diff
@@ -0,0 +1,43 @@
+[rhel-7-server-os-rpms]
+metadata_expire = 86400
+baseurl = https://pulp04.web.prod.ext.phx2.redhat.com/content/dist/rhel/server/7/$releasever/$basearch/os
```
pszubiak: It seems that we reveal internal URL here. Probably we should avoid this. Maybe we can inject this as a variable?

jenkins-slaves/ansible/kie-rhel7.yaml
```diff
@@ -0,0 +1,364 @@
+---
+- hosts: packer
+  remote_user: root
+  gather_facts: false
+  vars :
+    maven_versions: [3.0.5, 3.2.3, 3.2.5, 3.3.9, 3.5.0, 3.5.2]
+    jdk_download_urls: [
+      "https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/slaves-resources/lastSuccessfulBuild/artifact/jdk-6u45-linux-x64.tar.gz",
```
pszubiak: It seems that here we have internal URL too.

jenkins-slaves/ansible/kie-rhel7.yaml
```diff
@@ -0,0 +1,364 @@
...
+  - name: Update all packages
+    yum:
+      name: "*"
+      state: latest
+
+  - name: Add user jenkins
+    user:
+      name: jenkins
+      # droolsjbpm.ci1
+      password: $6$rounds=656000$skw/fVgypJZWP8m1$04UK6Ax4.2hhJfcCbgGegNGl0brQtufYJd4lGn4Lw2RnZ8GQL50crLAd/PIU/.JC047TLybYfcdLL/27hkwlx0
```
pszubiak: The problem I see here is that salt will prevent only for rainbow table attacks. If this is a dictionary based password it still could be "recovered" without enormous computer power.

jenkins-slaves/ansible/data/config/pulp/pulp.repo
psiroky: I personally don't think this is an issue. But of course it can be replaced by variable.

jenkins-slaves/ansible/kie-rhel7.yaml
psiroky: The only reason the password is not in a clear-text is that ansible does not like that (the password is actually in a clear text just line above that). If we are concerned about security, then I would disable password login completely and just rely on the ssh keys.

jenkins-slaves/ansible/data/config/pulp/pulp.repo
pszubiak: For me personally it's not an issue too however I know that CE team is very serious about it. I don't know if this is policy or just CE team practice.

I think we can live with that. And if there is a sense in changing this maybe it's a good task for us to get some practice with the procedure?

jenkins-slaves/ansible/kie-rhel7.yaml
pszubiak: I don't know if we should be concerned about security. I probably pay too much attention to that. The problem is that addressing security concerns would decrease convenience of usage but on the other hand if we don't see security risk it doesn't mean that it doesn't exist. It's always trade off.

I would leave it as it is if people use this access method. If it's rarely used I would probably remove it.

psiroky:
Thanks for the review @pszubiak. I'll merge this as is and we can then create smaller tasks that you (or others) can take a look and fix.

PR #101, opened 2017-11-10
Build with Maven 3.5.2

psiroky:
Needed as a fix for https://issues.apache.org/jira/browse/MDEPLOY-221


mbiarnes:
should also the dailyBuild jobs and release-jobs on master should update to this Maven version?

psiroky:
I thought they are reusing the constants? Technically we only need this for the deploy jobs. Not sure if there is any harm upgrading it everywhere, probably not.

psiroky:
Let's upgrade only the deploy + PR jobs now. If everything goes well, we can upgrade the rest of the places.

PR #104, opened 2017-11-14
adapted to new and modyfied scripts

mbiarnes:
- adapted job to work with new scripts


mbiarnes:
MERGE together with https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/611

PR #119, opened 2017-12-04
apdated to use new update-versions-all.sh script

mbiarnes:
Merge only after https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/637/files as this groovy file needs the updated update-versions-all.sh script which includes the version upgrade of appformer and kie-soup.


mbiarnes:
@psiroky would you be so kind and merge this once approved?

PR #123, opened 2017-12-05
disabled jobs

mbiarnes:


pszubiak:
Wonder if we still need 7.4.x in case when fix for 7.0 LA would be needed?

mbiarnes:
@pszubiak I thought the 7.0.0.LA will be done from 7.3.x ?

pszubiak:
@mbiarnes Yes you are right. Sorry for confusion.

PR #138, opened 2018-01-26
removed uberfire git org/repo lines

mareknovotny:
I saw reference to Uberfire repo and trying to build in CI job so removing here for ensuring it is gone https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/droolsjbpm-build-bootstrap-pullrequests/735/

### DIFF
--- a/jenkins-slaves/ansible/kie-rhel7.yaml
+++ b/jenkins-slaves/ansible/kie-rhel7.yaml
@@ -308,8 +308,6 @@
     shell: cd /home/jenkins/git-repos && git clone --bare https://github.com/{{ item }}
     with_items:
      - errai/errai
-     - uberfire/uberfire
-     - uberfire/uberfire-extensions
      - dashbuilder/dashbuilder
      - kiegroup/kie-soup
      - kiegroup/appformer


mareknovotny:
@ryanzhang and @etirelli  I am trying to fix test environment, but not sure if this is only place to fix. We we will see after merging.

mareknovotny:
Isn't that automatic when merged on SCM change?

pszubiak:
Nope. It is also not verified. Verification is only for DSL. Is there any issue with test env?

mareknovotny:
I noticed that in https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/675 test failing on empty uberfire cloned repo directory

mbiarnes:
@mareknovotny @pszubiak I also see that dashbuilder/dashbuilder is still in this list - uberfire and dashbuilder are now under appformer - right?

pszubiak:
Yep. You are right! I have one concern about this. If we remove uberfire and dashbuilder repositories will 6.5.x PRs and daily build works? These builds still use uberfire and dashbuilder.

mareknovotny:
@mbiarnes that is true that dashbuilder was moved appformer repository, but here I was trying to fix the issue with nonexistent repository content. Even dashbuilder were moved the old repository is still with a real content even outdated so it has a place in git cache for old like 6.x PRs or CI jobs IMHO.

PR #139, opened 2018-02-01
Add kiegroup/drlx-parser repository to git cache.

pszubiak:


pszubiak:
@mareknovotny sounds like a very interesting idea. Are thinking about hook that would send a notification message when repository-list.txt file is updated?

mareknovotny:
@pszubiak sort of or just have one source of repositories?

pszubiak:
@mareknovotny one source of repositories sounds like the best solution to me. But don't have easy solution how to the problem that we have multiple branches with different repository lists and only one type of slave node. Do you have idea how to handle this?

mareknovotny:
I already thought about having more slave types, but not really sure it will be required as git cache is just support for reducing time of build initialization. It would not be the issue for 1-2 git repositories are not in cache.

PR #143, opened 2018-02-07
added optashift-employee-rostering

mbiarnes:


job-dsls/jobs/deploy_jobs.groovy
```diff
@@ -138,8 +138,12 @@ def final REPO_CONFIGS = [
         "jbpm-work-items"           : [
                 label      : "linux && mem4g",
                 timeoutMins: 30,
-                ircNotificationChannels: ["#jbpmdev"]
-        ]
+                ircNotificationChannels: ["#jbpmdev"],
+                downstreamRepos        : ["optashift-employee-rostering"]
```
ge0ffrey: optashift-employee-rostering is a downstream repo of optaplanner, not of jbpm-work-items?

ge0ffrey: Ah, but I get it, it is also a downstream repo of errai, so this is the first repo where optaplanner and errai are downstream of.

PR #145, opened 2018-02-08
added jobs for lienzo

mbiarnes:


mbiarnes:
will close this as a new PR with more commits will be raised

PR #151, opened 2018-02-12
Enable sourcemaps profile in downstream PR jobs

jhrcek:
After discussion with @manstis we agreed that it would be very useful for QE if kie-wb.war, which is built in downstream PR jobs, would have sourcemaps included. That basically means that java source code would be added to the final war file, making client-side code debugging in the browser possible. 

Addition of sourcemaps was implemented in separate maven profile in [this PR](https://github.com/kiegroup/kie-wb-distributions/pull/678) and merged both on master and 7.5.x

The intention of this PR is to make sourcemaps appear in kie-wb war which is archived as a part of full downstream builds (and nowhere else). @mbiarnes can you please check if I added the profile to correct place and if yes, re-generate jenkins jobs after this is merged?


jhrcek:
@mbiarnes If ok, can you merge this, please? I'm not sure who's the gatekeeper for this repo.. Also could you let me know when this new config is available in donwstream build jobs so I can test this?

PR #163, opened 2018-03-06
Prod errai

mbiarnes:
- a new DSL job is required for a temp. errai version (only for prod) like in daily builds
- the ZANATA version for Zanata projects changed to 7.7.0
- the jdk is defined as variable $javadk in DSJ jobs

IMPORTANT: please merge also https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/718

### DIFF
--- a/job-dsls/jobs/kie_release_jobs.groovy
+++ b/job-dsls/jobs/kie_release_jobs.groovy
@@ -57,6 +57,68 @@ def copyBinariesToFilemgmt="""
 sh \$WORKSPACE/scripts/droolsjbpm-build-bootstrap/script/release/kie-copyBinariesToFilemgmt.sh
 """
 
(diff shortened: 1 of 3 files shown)


mareknovotny:
This prod Errai release/job is only because we do not have upstream release for Errai we need, right?

mbiarnes:
@mareknovotny since errai does not form part of kiegroup it has its own release cycle. In daily Builds we clone the release (masetr or 7.5.x) and create a temp version (i.e. 4.2.0.20180206-008009). In community releases we use a fixed and closed release (4.1.3.Final) - but this doesn't have the commit for wildfly11. So for next community release we need a new version of errai. For prod. we do a temp version so we can tag and psh the tag to Gerrit.

PR #165, opened 2018-03-08
Adds 8GB swap file.

pszubiak:

### DIFF
--- a/jenkins-slaves/ansible/kie-rhel7.yaml
+++ b/jenkins-slaves/ansible/kie-rhel7.yaml
@@ -357,6 +357,28 @@
       group: jenkins
       mode: 0755
 
+# Add swap file
+  - name: Create swap file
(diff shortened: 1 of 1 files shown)


mareknovotny:
8GB swap and what is the memory size dedicated ? I don't recall if that was higher than 8GB.

pszubiak:
It depends we have several templates:

ci-rhos-rhel7-large - 8 GB

ci-rhos-rhel7-medium - 4GB

ci-rhos-rhel7-xlarge - 24GB

mareknovotny:
so for *xlarge that swap size would be not enough imho

pszubiak:
Yep you can be right. I'm testing that now.

AntStephenson:
8GB will probably be OK, but it'll be the _minimum_ we'd want to allocate.

PR #167, opened 2018-03-13
fixed additional tests

mbiarnes:


job-dsls/jobs/FlowJob.groovy
```diff
@@ -739,7 +739,7 @@ job("kie-docker-ci-images-${kieMainBranch}") {
     description("Builds CI Docker images for master branch. <br> IMPORTANT: Created automatically by Jenkins job DSL plugin. Do not edit manually! The changes will get lost next time the job is generated. ")
 
     parameters {
-        stringParam("kieVersion", "8.0.0-SNAPSHOT", "Please edit the version of the kie release <br> i.e. typically <b> major.minor.micro.EXT </b>i.e. 8.0.0.Beta1<br> Normally the kie version will be supplied by parent job <br> ******************************************************** <br> ")
+        stringParam("kieVersion", "7.7.0-SNAPSHOT", "Please edit the version of the kie release <br> i.e. typically <b> major.minor.micro.EXT </b>i.e. 7.7.0.Beta1<br> Normally the kie version will be supplied by parent job <br> ******************************************************** <br> ")
```
pszubiak: Is 7.7.0-SNAPSHOT fine for 7.5.x branch?

mareknovotny: actually it is better than 8.0.0-SNAPSHOT, which doesn'ŧ exist or only by mistake I think

job-dsls/jobs/FlowJob.groovy
```diff
@@ -451,9 +451,9 @@ matrixJob("jbpmTestContainerMatrix-kieAllBuild-${kieMainBranch}") {
     }
 
     publishers {
-        wsCleanup()
-        archiveJunit("**/TEST-*.xml")
+        archiveJunit("**/target/failsafe-reports/TEST-*.xml")
```
mareknovotny: I think this notation is the same, not sure why it should not be ;)

mbiarnes: it is - it is - but I like to have it all the same notation :)

job-dsls/jobs/FlowJob.groovy
mbiarnes: nop - but this is only a text in the description - this value comes via sh

PR #186, opened 2018-03-27
Fix kie-wb-distributions oom error.

pszubiak:


pszubiak:
@mareknovotny Yes there was time when mem16g label was in fact 24GB of RAM template. After adding SWAP it was reverted and new mem24g template was added. It seems that now kie-wb-distributions with -T2 needs more than 24GB of RAM + SWAP to build.

PR #194, opened 2018-03-29
fixed parameter definition and removed timeout

mbiarnes:
have to use double square bracket for variable definition and removed time outs in builds as they are to small depending the traffic on nodes


pszubiak:
Wonder if removing time-out from daily build is a good move. Why do we want to do that?

mbiarnes:
because sometimes, when Jenkins is busy, it could be that a job has to wait in the queye before it runs - this seems to be in the timeout - we have a lot breaks of jobs cause of timeouts. the timeout for build was set to 340 m = 5 hr 40 min ! so if you think it is better to increase the timeout value  - we can do this.

mbiarnes:
timeout increased

mbiarnes:
Jenkins please retest this

PR #196, opened 2018-03-29
fixed parameter definition and removed timeout

mbiarnes:
have to use double square bracket for variable definition and removed time outs in builds as they are to small depending the traffic on nodes


mbiarnes:
Jenkins please retest this

PR #199, opened 2018-04-13
added new profile

mbiarnes:


mbiarnes:
@pszubiak don't merge before next Monday

PR #216, opened 2018-05-16
Unpack qa nexus

mbiarnes:
applied all changes for "deploying" (upload a zip to QA Nexus)
- new settings.xml
- changed hard coded properties to use of variables
- adapted to higher firefox driver (as it is in other scripts) 
- use of credentialsBinding (so the user and pwd is not visible since stored on Jenkins)
- fixed label for errai build
- adjusted additional tests to grab the sources from the right URL

These changes are now for master only. Lets see if tomorrow everything builds properly and we will do these changes also to 7.7.x and 7.5.x.

### DIFF
--- a/job-dsls/jobs/kie_dailyBuild_pipeline.groovy
+++ b/job-dsls/jobs/kie_dailyBuild_pipeline.groovy
@@ -158,25 +158,26 @@ git checkout -b $erraiVersionNew $erraiBranch
 sh updateVersions.sh $erraiVersionOld $erraiVersionNew
 # build the repos & deploy into local dir (will be later copied into staging repo)
 deployDir=$WORKSPACE/deploy-dir
-# (1) do a full build, but deploy only into local dir
+# do a full build, but deploy only into local dir
 # we will deploy into remote staging repo only once the whole build passed (to save time and bandwith)
 mvn -U -B -e clean deploy -T2 -Dfull -Drelease -DaltDeploymentRepository=local::default::file://$deployDir -s $SETTINGS_XML_FILE\\
  -Dmaven.test.failure.ignore=true -Dgwt.compiler.localWorkers=3
-# (2) upload the content to remote staging repo
+# unpack zip to QA Nexus
 cd $deployDir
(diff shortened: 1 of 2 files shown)


job-dsls/jobs/kie_dailyBuild_pipeline.groovy
```diff
@@ -276,15 +280,17 @@ rm -rf \\`find \\$baseDir -type f -name "*zip"\\`
...
  -Dkie.maven.settings.custom=$SETTINGS_XML_FILE -Dmaven.test.redirectTestOutputToFile=true -Dmaven.test.failure.ignore=true\\
  -Dgwt.memory.settings="-Xms1g -Xmx10g -Xms1g -Xss1M" --clean-up-script="$WORKSPACE/clean-up.sh"
 
-# (2) upload the content to remote staging repo
-mvn -B -e org.sonatype.plugins:nexus-staging-maven-plugin:1.6.8:deploy-staged-repository -DnexusUrl=https://repository.jboss.org/nexus -DserverId=jboss-releases-repository\\
- -DrepositoryDirectory=$deployDir -s $SETTINGS_XML_FILE -DstagingProfileId=15c3321d12936e -DstagingDescription="kie $kieVersion" -DkeepStagingRepositoryOnCloseRuleFailure=true -DstagingProgressTimeoutMinutes=80
+# unpack zip to QA Nexus
+cd $deployDir
+zip -r kiegroup .
+curl --upload-file kiegroup.zip -u $kieUnpack -v http://bxms-qe.rhev-ci-vms.eng.rdu2.redhat.com:8081/nexus/service/local/repositories/kieAllBuild/content-compressed
```
winklerm: Afaik, the Jenkins URL redirects to https always - so you need to keep an eye on SSL certificate expiration even though you are not using https explicitely. Just to let you know..

winklerm: Sorry, my mistake - I confused our Nexus instance with our Jenkins master, please disregard my previous comment.

PR #219, opened 2018-05-18
timer set taking in account the time diff

mbiarnes:


job-dsls/jobs/kie_dailyBuild_pipeline.groovy
```diff
@@ -125,7 +125,7 @@ pipelineJob("kieAllBuildPipeline-${kieMainBranch}") {
     }
 
     triggers {
-        cron("H 2 * * *")
+        cron("H 22 * * *")
```
mareknovotny: so we changed the server datetime or time zone? I barely noticed that you were solving different trigger time on updated Jenkins instance, right?

pszubiak: @mareknovotny @mbiarnes Yes there is an issue with scheduled trigger on updated Jenkins. It works after restart but after some time it stuck. During investigation of this issue we noticed that UTC TZ is used by Jenkins and there is a shift CEST and Jenkins default TZ.

mareknovotny: ok, so it uses UTC or some other time zone?

pszubiak: It use UTC. Maybe a good solution is to specify TZ explicit in cron string instead of doing calculation?

mareknovotny: I think it doesn't matter to use timezone or not, just be consistent in all job configs

PR #247, opened 2018-06-28
fixed path to test result directory

mbiarnes:


mareknovotny:
I have very the same question, as I suppose that all such test modules should have it or do they generate something else as test results/reports?

mbiarnes:
@mareknovotny @winklerm maybe this allow empty test results is obsolete?
Very often these test failed, not because of the build - this was successful - but because test results were not found. The directories were filesafe-reposrts or surefire-reports, so I think with *-resports/ it is OK. If you want I can remove the allow empty.

jhrcek:
@mbiarnes from your description it seems to me that we should fix the incorrect path to test reports instead of "ignoring" the problem by allowing empty test results. If this kind of failures happened recently, could you please provide link(s) to a couple of builds where this happen, so we can decide if it's ok that test results were not found?

mareknovotny:
+1 to @jhrcek 's comment, sometimes it is not good to ignore a random failure or bad configuration. What is the case for this?

mbiarnes:
@jhrcek @mareknovotny the "allowEmptyResults" was removed

PR #259, opened 2018-07-10
changed absolute timeout to elastic timeout

mbiarnes:


jhrcek:
This looks reasonable. Just for myself and for reference:

`elastic(X,Y,Z)` means that build will timeout when it take X/100 times longer than the reference build duration of the last Y builds and will use Z minutes timeout when no successful builds available as reference

mbiarnes:
@jhrcek correct

PR #293, opened 2018-08-09
added variable in target for settings.xml

mbiarnes:


mbiarnes:
@pszubiak absolutely - but this will be next step after migration - clean the scripts from duplication and use templates - this is on my ToDo list

PR #321, opened 2018-09-06
adjusted PR job for kie-docs

mbiarnes:


job-dsls/jobs/pr_jobs.groovy
```diff
@@ -87,6 +87,14 @@ def final REPO_CONFIGS = [
         "jbpm-wb"                   : [
                 label: "kie-rhel7 && kie-mem16g"
         ],
+        "kie-docs"                   : [
+                label: "kie-rhel7 && kie-mem16g",
+                artifactsToArchive: DEFAULTS["artifactsToArchive"] + [
+                        "doc-content/drools-docs/target/generated-docs/html_single/**/*",
+                        "doc-content/jbpm-docs/target/generated-docs/html_single/**/*",
+                        "doc-content/optaplanner-wb-es-docs/target/generated-docs/html_single/**/*"
```
mareknovotny: we could limit the path to only one with starting ** or not? @jomarko is there a reason to have listed all 3 there?

`**/target/generated-docs/html_single/**/*"`

pszubiak:
And PR job for kie-docs is now standalone (not in main project build) to avoid upstream and downstream repos build. The definition of job is here:

https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/jobs/kie_docs_pr.groovy

job-dsls/jobs/pr_jobs.groovy
mbiarnes: @mareknovotny if all needed docs are in 
**/target/generated-docs/...
it makes sense to have only one line

mbiarnes:
@pszubiak so - best to remove it again?

pszubiak:
@mbiarnes @mareknovotny When kie-docs job was switched to standalone I didn't make archive artifacts functionality. I should add a method to parent class [1] PrVerificationJob to add functionality with archiving artifacts. So far haven't found time for this.

[1] https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/src/main/groovy/org/kie/jenkins/jobdsl/templates/PrVerificationJob.groovy

mbiarnes:
@mareknovotny @jomarko @pszubiak 
I will close this PR since this is not valid anymore. Will work with Piotr on another.

pszubiak:
@mareknovotny Actually @mbiarnes foud that I did archive functionality as an ugly if ;-)

https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/jobs/pr_jobs.groovy#L184-L193

pszubiak:
@mbiarnes @mareknovotny Sorry it's not a parent class. My brain is not working today.. I need more coffee.



There is no code about archiving in PrVerificationJob class. We can copy that if or create a new method.

job-dsls/jobs/pr_jobs.groovy
jomarko: Yes sure, one string is fine, I was just not sure about wildcards so this was first thing I tried. Thanks for looking into it, both!

PR #325, opened 2018-09-10
BXMSPROD-69: Archive kie-docs artifacts

jomarko:
@mbiarnes @pszubiak would you mind to review?


job-dsls/src/main/groovy/org/kie/jenkins/jobdsl/templates/PrVerificationJob.groovy
```diff
@@ -197,4 +203,35 @@ class PrVerificationJob extends BasicJob {
             }
         }
     }
+
```
pszubiak: Thank you for PR! Do we need this?

jomarko: Thanks for spotting, I will remove.

jomarko:
@pszubiak I updated the PR, please have a look and let me know.

jomarko:
@pszubiak I named the argument to be consistent as discussed on slack

PR #326, opened 2018-09-10
upgraded Maven min version to 3.5.2

mbiarnes:
Please do not merge before: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/825 was merged

### DIFF
--- a/job-dsls/jobs/kie_dailyBuild_pipeline.groovy
+++ b/job-dsls/jobs/kie_dailyBuild_pipeline.groovy
@@ -4,8 +4,8 @@ import org.kie.jenkins.jobdsl.Constants
 
 def javadk=Constants.JDK_VERSION
 def javaToolEnv="KIE_JDK1_8"
-def mvnToolEnv="KIE_MAVEN_3_3_9"
-def mvnVersion="kie-maven-3.3.9"
+def mvnToolEnv="KIE_MAVEN_3_5_2"
+def mvnVersion="kie-maven-3.5.2"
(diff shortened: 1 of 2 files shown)


mareknovotny:
why don't we upgrade to the latest available 3.5.4?

pszubiak:
@mareknovotny We need to modify and rebuild slave nodes for that [1]. Let's do this just after 7.1 is out.

[1] https://github.com/kiegroup/kie-jenkins-scripts/blob/master/jenkins-slaves/ansible/kie-rhel7.yaml#L6

mareknovotny:
@pszubiak so even this upgrade to 3.5.2 should be postponed, right?

pszubiak:
@mareknovotny we don't have to postpone it. Maven 3.5.2 is already installed on slave nodes. In fact, PR jobs are using Maven 3.5.2 and the rest of the jobs are using 3.3.9. So it's good to unify this.

mbiarnes:
@mareknovotny is the upgrade for 3.5.2 OK for you? Nothing else had to be done.

pszubiak:
@mareknovotny @mbiarnes Wonder if we can merge this now as this PR will clean up maven version usage that we use for our builds in master.

Most of jobs (PR, dowstream and deploy) use Maven 3.5.2 defined in Constants [1]. Rest of jobs is using Maven 3.3.9.

We can bump version of maven to 3.5.4 in other PR. I think we should aslo unify maven version definition across all our jobs, I don't see reason to define it in so many places. What do you think?

PR #327, opened 2018-09-12
Update firefox to latest ESR versions

jhrcek:
@mbiarnes Selenium tests in PR jobs started to fail recently, because it takes too long to start firefox before tests. This is the first attempt at stabilization: I updated to latest available firefox ESR (Extended Support Release) versions. Also replaced one usage of (quite old) firefox 38 and replaced it with 45. 

As a followup to this PR I will update to latest version of selenium in the next sprint, so we can star using even more recent version of firefox (52 and 60) which currently used version of selenium doesn't yet support.


mbiarnes:
@jhrcek @domhanak the changes will be available when the nodes are re-done - it is planned to do this after 7.1.0.CR1 is out

PR #239, opened 2018-06-22
Disable -T option in in full downstream / PR builder jobs

jhrcek:
As agreed on F2F, I'm disabling `-T` maven option in full downstream / PR builder jobs.
It seems to cause trouble (test instability) without saving time (multiple reruns needed).
This is expected to slow down most individual builds, but speedup time to merge (less rebuilds needed).
To make this change data-driven, I collected build duration of all PR builder and downstream-pullrequests jobs together with basic stats (n = number of builds used for measurement, σ = standard deviation of build time, avg=average duration in HH:MM:SS).
Later I'll use that data to determine how much exactly the build was slowed down.

The data are based on all available build history or [PR jobs](https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/PRs/).
The data only takes into consideration jobs with status SUCCESS / UNSTABLE (= everything was build with some potential test failures) and ignores all jobs with status FAILURE / ABORTED (= infra failures, build timeouts etc.).
Some `pullrequests` jobs don't have associated `downstream-pullrequests` job (marked with NO JOB).
Some jobs didn't have any successful / unstable builds in history (marked NO BUILDS)/

Measurements done on 22th June 2018

| repository | pullrequests | downstream-pullrequests |
|------------|--------------|-------------------------|
| appformer | avg=00:23:06 ,  σ= 00:00:58, n=3 | avg=04:14:27 , σ=00:28:26, n=2 |
| bxms-patch-tools | avg=00:14:04 , σ=00:12:12, n=8 | NO JOB |
| drlx-parser | avg=00:33:24 , σ=00:35:21, n=2 | NO BUILDS |
| drools | avg=00:30:43 , σ=00:01:55, n=5 | avg=04:12:31 , σ=00:50:18, n=2 |
| drools-wb | avg=00:35:50 , σ=00:00:53, n=3 | avg=02:01:30 , σ=00:00:00, n=1 |
| drools-website | avg=00:00:07 , σ=00:00:03, n=2 | NO JOB |
| droolsjbpm-build-bootstrap | avg=00:06:09 , σ=00:01:07, n=4 | avg=04:21:43 , σ=00:53:16, n=2 |
| droolsjbpm-integration | avg=01:44:30 , σ=00:08:21, n=15 | avg=02:17:51 , σ=00:15:18, n=2 |
| droolsjbpm-knowledge | avg=00:07:04 , σ=00:00:00, n=1 | avg=05:15:17 , σ=00:00:00, n=1|
[...]

PR #328, opened 2018-09-12
disable parallel build on deploy jobs

jsoltes:
For consistency, enforcing the same behavior on deploy jobs as was set to the pr jobs, see https://github.com/kiegroup/kie-jenkins-scripts/pull/239.

@pszubiak @mbiarnes can you please review?


pszubiak:
@mbiarnes If people are using test results from deploy jobs it makes sense to disable parallel maven builds. Can you take a look?

jhrcek:
Just FYI, this is not about using test results, but about build consistency. We know there were some failures caused by the parallel build. Disabling it will prevent mysterious failures that would happen in deploy builds only, but not in other builds.

pszubiak:
@jhrcek @mbiarnes @jsoltes Yes we hit build failures caused by parallel build too. In our case, it was the kie-wb-distribution component and the reason was a significantly larger amount of memory used by a parallel build.

+1 to this change.

However, it's always a trade-off. In case of kie-wb-distribution disabling -T2 increase build time. The measurement was done about half a year ago, and it changes on average from 33 min to 49 min.

PR #340, opened 2018-09-30
Set webdriver.firefox.bin for downstream repos

jhrcek:
When updating to latest selenium I made it mandatory to specify
Firefox binary path (to avoid using firefox provided by OS).
https://github.com/kiegroup/kie-wb-distributions/blob/master/kie-wb-tests/kie-wb-tests-gui/pom.xml#L21-L22

That lead to failures in full downstream builds, because that property wasn't specified.


jhrcek:
Merging this because the fact that this property is not defined breaks the selenium tests in all full downstream build jobs.

PR #344, opened 2018-10-10
added new groovy script

mbiarnes:
for the compilation of all downstream reps skipping tests and skipping gwt compilation

### DIFF
--- a/job-dsls/jobs/compile_downstream_build.groovy
+++ b/job-dsls/jobs/compile_downstream_build.groovy
@@ -1,0 +1,204 @@
+/**
+ * Creates downstream pullrequest (PR) jobs for kiegroup GitHub org. units.
+ * These jobs execute the downstream build skipping tests and skipping the gwt compilation
+ * for a specific PR to make sure the changes do not break the downstream repos.
+ */
+
+def final DEFAULTS = [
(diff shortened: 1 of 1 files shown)


baldimir:
Looks ok, but agree with @pszubiak that if there is a similar code, it should be refactored, because duplicate code is a pain to maintain.

mbiarnes:
@psiroky @baldimir you're right - but I don't know how to refactor it. What we need now is a fast solution because a lot of people is complaining. For sure I will need some days to find a way how to "refactor" the code.

mbiarnes:
@pszubiak @jhrcek @winklerm @baldimir 
did all the changes for master - lets see if this works
the seed job is https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/DSL-compile-downstream-master/ - the new jobs will be on the "old" Jenkins instance until we have solved our Jenkins CI errors - this is an absolutely temporary solution.
If this works, we have to cherry-pick this for 7.11.x too.

jhrcek:
@mbiarnes when the jobs for this are generated, could you please send an email to bsig explaining how these newly created compile jobs can be used (e.g. what github comment to use to trigger the build etc)?

PR #345, opened 2018-10-10
[7.11.x] added new groovy script

mbiarnes:
(cherry picked from commit 3003755b8a543f7d2001b0864bff2d97c93d69e8)

for the compilation of all downstream reps skipping tests and skipping gwt compilation

### DIFF
--- a/job-dsls/jobs/compile_downstream_build.groovy
+++ b/job-dsls/jobs/compile_downstream_build.groovy
@@ -1,0 +1,206 @@
+/**
+ * Creates downstream pullrequest (PR) jobs for kiegroup GitHub org. units.
+ * These jobs execute the downstream build skipping tests and skipping the gwt compilation
+ * for a specific PR to make sure the changes do not break the downstream repos.
+ */
+import org.kie.jenkins.jobdsl.Constants
+
+def final DEFAULTS = [
+        ghOrgUnit              : Constants.GITHUB_ORG_UNIT,
+        branch                 : Constants.BRANCH,
+        timeoutMins            : 600,
+        label                  : "kie-rhel7 && kie-mem8g",
+        ghAuthTokenId          : Constants.GITHUB_AUTH_TOKEN,
+        upstreamMvnArgs        : "-B -e -T1C -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dfindbugs.skip=true -Drevapi.skip=true clean install",
+        downstreamMvnGoals     : "-B -e -nsu -fae clean install -Dfull=true -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true",
+        artifactsToArchive     : [
+                "**/target/*.log",
+                "**/target/testStatusListener*",
+                "**/target/screenshots/**",
+                "**/target/kie-wb*wildfly*.war",
+                "**/target/kie-wb*eap*.war",
+                "**/target/kie-drools-wb*wildfly*.war",
+                "**/target/kie-drools-wb*eap*.war",
+                "**/target/kie-server-*ee6.war",
+                "**/target/kie-server-*ee7.war",
+                "**/target/kie-server-*webc.war",
+                "**/target/jbpm-server*dist*.zip"
+        ],
+        excludedArtifacts      : [
+                "**/target/checkstyle.log"
+        ]
+]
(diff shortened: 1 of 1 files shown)


baldimir:
Looks ok, but agree with @pszubiak that if there is a similar code, it should be refactored, because duplicate code is a pain to maintain.

mbiarnes:
@psiroky @baldimir you're right - but I don't know how to refactor it. What we need now is a fast solution because a lot of people is complaining. For sure I will need some days to find a way how to "refactor" the code.

job-dsls/jobs/compile_downstream_build.groovy
```diff
@@ -0,0 +1,206 @@
...
+        "droolsjbpm-integration"    : [],
+        //"droolsjbpm-tools"          : [], // no other repo depends on droolsjbpm-tools
+        "kie-uberfire-extensions"   : [],
+        "kie-wb-playground"         : [],
+        "kie-wb-common"             : [],
+        "drools-wb"                 : [],
+        "optaplanner-wb"            : [],
+        "jbpm-designer"             : [],
+        "jbpm-wb"                   : []
+        //"kie-wb-distributions"      : [] // kie-wb-distributions is the last repo in the chain
```
winklerm: Just out of curiosity, why is `kie-wb-distributions` commented out?

mbiarnes: @winklerm good question - this was taken 1:1 from Petr - so we did not change it. There should be no other rep dependent of kie-wb-distributions? (like droolsjbpm-tools)

job-dsls/jobs/compile_downstream_build.groovy
```diff
@@ -0,0 +1,206 @@
...
+        downstreamMvnGoals     : "-B -e -nsu -fae clean install -Dfull=true -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true",
+        artifactsToArchive     : [
+                "**/target/*.log",
+                "**/target/testStatusListener*",
+                "**/target/screenshots/**",
+                "**/target/kie-wb*wildfly*.war",
+                "**/target/kie-wb*eap*.war",
+                "**/target/kie-drools-wb*wildfly*.war",
+                "**/target/kie-drools-wb*eap*.war",
+                "**/target/kie-server-*ee6.war",
```
jhrcek: @mbiarnes you told me that kie-server ee6 is no longer being generated, so it should probably be removed from this "what to archive" section.

job-dsls/jobs/compile_downstream_build.groovy
```diff
@@ -0,0 +1,206 @@
...
+import org.kie.jenkins.jobdsl.Constants
+
+def final DEFAULTS = [
+        ghOrgUnit              : Constants.GITHUB_ORG_UNIT,
+        branch                 : Constants.BRANCH,
+        timeoutMins            : 600,
+        label                  : "kie-rhel7 && kie-mem8g",
+        ghAuthTokenId          : Constants.GITHUB_AUTH_TOKEN,
+        upstreamMvnArgs        : "-B -e -T1C -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dfindbugs.skip=true -Drevapi.skip=true clean install",
+        downstreamMvnGoals     : "-B -e -nsu -fae clean install -Dfull=true -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true",
```
jhrcek: Since we won't be running any tests in these "just compile" jobs, I would suggest adding -T1C into the downstreamMvnGoals - here maven parallelism shouldn't hurt, on the contrary it will make the jobs run faster.

job-dsls/jobs/compile_downstream_build.groovy
jhrcek: @winklerm To add my understanding: I believe it's ok to have `kie-wb-distributions` and `droolsjbpm-tools` commented out. 
The purpose of this groovy script is to generate "compile downstream" jobs is to make sure that changes in repo X don't break any of the repos that depend on X. In this particualr case there are no repos that depend on kie-wb-distributions and on droolsjbpm-tools, so we don't need jobs for those.

job-dsls/jobs/compile_downstream_build.groovy
pszubiak: +1; adding -T1C seems to be a very good idea.

PR #348, opened 2018-10-14
Don't remove JUnit reports when cleaning up build artifacts

jhrcek:
My previous change broke the kieAllBuild, because the cleanup also removed the junit reports, that the post build step tries to archive, resulting in failure:

```
ERROR: Step ‘Publish JUnit test result report’ failed: No test report files were found. Configuration error?
```

This change tries to rectify this by adding the test reports to the index, preventing their removal in the `git clean` step.


jhrcek:
@mbiarnes FYI I'm going to merge this to fix the daily build asap.

PR #349, opened 2018-10-14
Speed up the PR builder jobs

mareknovotny:
This is an experimental thing for our PR builders.

There a real concerns from @krisv to have upstream whole chain build before testing every PR. As that was originally just to have up to date the SNAPSHOTs of all repos for PR, we can try to rely on SNAPSHOTs available from the last daily build and just test the PR changes as in most cases it could be ok to build only that change. Otherwise we do full downstream build.

We decided to try out for 1 week until some big issue don't show up.

WDYT @mbiarnes, @pszubiak?

### DIFF
--- a/job-dsls/jobs/pr_jobs.groovy
+++ b/job-dsls/jobs/pr_jobs.groovy
@@ -210,7 +210,7 @@ for (repoConfig in REPO_CONFIGS) {
         }
 
         steps {
-            if (repo != "jbpm-work-items") {
+            /* if (repo != "jbpm-work-items") {
                 configure { project ->
                     project / 'builders' << 'org.kie.jenkinsci.plugins.kieprbuildshelper.UpstreamReposBuilder' {
                         mavenBuildConfig {
@@ -220,7 +220,7 @@ for (repoConfig in REPO_CONFIGS) {
                         }
                     }
                 }
-            }
+            }*/
             maven {
                 mavenInstallation("kie-maven-${Constants.MAVEN_VERSION}")
                 mavenOpts("-Xms1g -Xmx3g -XX:+CMSClassUnloadingEnabled")


krisv:
@mareknovotny I think there might be one possible use case that might cause issues here: assume we merge a multi-repo PR where the second repo depends on the changes of the first (let's say a drools + jbpm PR), then I guess as soon as it merged, it will cause issues because it will take until the next nightly build for the snapshots to be uploaded with the changes.  So all PRs on the second repo will fail until then.  Not sure how often that would occur though.  A way to publish new snapshots on demand for a specific repo might be useful here.

pszubiak:
If I understand correctly SNAPSHOTs are built "immediately" after a PR is merged to the repository by deploy job. So it won't be required to wait for a whole day.

jhrcek:
@mareknovotny @krisv I think there's another thing that it will break:
Imagine you're doing multi-repo PR and you do changes in 3 repos, e.g.:

kie-soup
appformer
kie-wb-common

When you make an API change in kie-soup, the change in this PR will cause the PR builds in appformer and kie-wb-common to fail, because the required downstream repos wouldn't be built.

jhrcek:
@pszubiak could you please explain what mechanism (which job or something) ensures that snapshots are built immediately? I'd like to understand this better.

pszubiak:
My understanding is that once PR is merged then deploy jobs is executed [1]. If deploy job build is successful it deploys SNAPSHOT artifact to nexus.

The nightly jobs don't build SNAPSHOTs. They build a version with a date. Like  7.13.0.20181014-031034.

[1] https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/deployedRep/

manstis:
As @jhrcek states not building related repositories when a PR spans multiple repositories is going to be a killer for a non insignificant amount of the UI development - which already probably has the slowest (if not one of the slowest) turn-over of PRs (from submission to green builds to merge). Can this feature not be controlled by a new Jenkins "switch" so PRs that span multiple repositories have the option of building all applicable repositories? IDK, like we have "ex3cute fu11 downstr3am bu1ld" (I don't want to trigger one here!) and "r3t3est th1s" etc could we have "ex3cute bu1ld across repositories" type command?

pszubiak:
@manstis This PR won't affect downstream builds, they will work as usual.

Technically it's possible to create a new label like "r3t3est th1s" however we are adding now a new label for compile downstream builds [1] and maintaining so many job types starting to be an issue.

@jhrcek @mareknovotny @mbiarnes Wonder if we should think more about it before merging. It looks to me that a solution could be building upstream repos only if upstream repo has a PR with the same Jira issue ID otherwise build only a repository where PR was raised. What do you think?

[1] https://github.com/kiegroup/kie-jenkins-scripts/pull/345

jhrcek:
> The other benefit is that we reduce usage of org.kie.jenkinsci.plugins.kieprbuildshelper.UpstreamReposBuilder plug-in with is not maintained since Petr left.

Until now I wasn't a plugin like this exists. What kinds of issues are there with this plugin? Shouldn't we rather fix them instead of reducing it's usage? It seems that Petr invested a lot of effort into making the PR builder jobs work, I think we shouldn't throw that away.

> Wonder if we should think more about it before merging. It looks to me that a solution could be building upstream repos only if upstream repo has a PR with the same Jira issue ID otherwise build only a repository where PR was raised. What do you think?

I'd consider this (only checking out and building repos with related PRs) as the only robust solution.

pszubiak:
> > The other benefit is that we reduce usage of org.kie.jenkinsci.plugins.kieprbuildshelper.UpstreamReposBuilder plug-in with is not maintained since Petr left.
> 
> Until now I wasn't a plugin like this exists. What kinds of issues are there with this plugin? Shouldn't we rather fix them instead of reducing it's usage? It seems that Petr invested a lot of effort into making the PR builder jobs work, I think we shouldn't throw that away.

@jhrcek +1 we shouldn't throw it away. Or if we want to get rid of custom plug-ins we should move its logic to Jenkins pipeline. So far this plug-in [1] is a foundation for our PR and downstream builds. We (Ant, Michael and myself) don't maintain it because don't have enough time to take care of it and it seems that other teams don't have spare time either.

> 
> > Wonder if we should think more about it before merging. It looks to me that a solution could be building upstream repos only if upstream repo has a PR with the same Jira issue ID otherwise build only a repository where PR was raised. What do you think?
> 
> I'd consider checking out and building repos with related PRs as the only robust solution.

Yes, at the first look, it seems that this functionality could be implemented in kieprbuildshelper plug-in.

[1] https://github.com/kiegroup/kie-build-helper-jenkins-plugin

mareknovotny:
so what will we do with the reducing the prebuilding?

pszubiak:
@mareknovotny I would vote for enhancing kie builder helper plugin.

mareknovotny:
@pszubiak what enhancement we would like to implement?
Even with enhanced or implementing in KIE BHP, we shoul drop for some types of build job(s) the prebuild like it is in this PR.

pszubiak:
@mareknovotny The idea was to pre-build upstream repository only if current and upstream repository PRs has the same Jira issue ID in the commit message. Otherwise, the plugin should skip upstream repository build. This solution wasn't discussed enough, what do you think about it?

mareknovotny:
@pszubiak correct, but there is 2 builds - one builds the whole chain without tests and then there is the PR only branches build. And we are trying to eliminate the first step.
GHPR plugin still does the work for the same branch name build in multiple repositories, or am I something missing?

kie-ci3:
Can one of the admins verify this PR? Comment with 'ok to test' to start the build.

PR #365, opened 2018-11-05
Rename profiles after wildfly14 upgrade

jhrcek:
Let's keep our CI consistent with the configuration in poms

Should be merged with:
https://github.com/kiegroup/kie-jenkins-scripts/pull/365
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/875
https://github.com/kiegroup/jbpm/pull/1366
https://github.com/kiegroup/droolsjbpm-integration/pull/1627
https://github.com/kiegroup/kie-wb-distributions/pull/841


rsynek:
@jhrcek why not renaming the profile to just "wildfly" and thus avoid renaming it again in every 9 months?

mswiderski:
@jhrcek I agree we should stick to wildly as the default profile of widely version support. We can then have (only if needed) additional profiles for other versions of wildfly that are named with version number. Same should apply to tomcat.

sutaakar:
Sounds good to me.

PR #397, opened 2019-01-23
start of pipeline delayed by 5 min to avoid the same suffix as master

mbiarnes:


job-dsls/jobs/kie_dailyBuild_pipeline.groovy
```diff
@@ -139,7 +139,7 @@ pipelineJob("${folderPath}/kieAllBuildPipeline-${kieMainBranch}") {
     }
 
     triggers {
-        cron("H 20 * * *")
+        cron("H 20 5 * *")
```
winklerm: According to Jenkins help, the cron expression should be rather something such as "10 20 * * *" - the H stands for "hash", the format is "MINUTE HOUR DOM MONTH DOW".

winklerm: The other branch daily build pipeline would need to have a different value then.

mbiarnes: * 20 * * * and 05 20 * * * ?

winklerm: Rather "0 20 * * *" and "05 20 * * *" - the first * would mean every minute.

mbiarnes: changed this online to H 19 * * * (master) and left H 20 * * * (7.14.x) - will see tomorrow!

winklerm: Sounds good to me!

PR #407, opened 2019-02-11
BXMSPROD-246 use different GitHub token for each job type.

pszubiak:


pszubiak:
Jenkins retest this

PR #422, opened 2019-02-21
disabled pipeline job

mbiarnes:


mbiarnes:
> What do you think about disabling PR, downstream and compile downstream jobs for 7.14.x too?

hmm - as it doesn't be triggered it doesn't hurt to be there "enabled" - I don't know if these branch is still in maintenance

PR #442, opened 2019-04-01
adapted script to use fixed errai version

mbiarnes:
MERGE this only after https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/943 and https://github.com/kiegroup/appformer/pull/673 was merged.


mareknovotny:
we could eliminate also all errai builds from daily and PR builder jobs when we have it set to fixed version.
Do you want to do it in this PR too or in the follow up PR?

mbiarnes:
follow up :) - if not I am afraid to loose the "overview"

mareknovotny:
ok

mbiarnes:
@mareknovotny btw. in dailyBuilds we won't have any more the errai build (fixed with this PR), the master PRs are copied manually - so they are not generated by an groovy script. Means the two PRs we have for errai we can ignore - or remove them. The next time we branch from master they are not generated and not copied. I think we can leave it now like it is.

PR #450, opened 2019-04-08
new Red Hat  UMB trigger difinition

mbiarnes:


mareknovotny:
how is this related to @baldimir and @MarianMacik work on pipeline?

baldimir:
@mareknovotny this is unrelated. From looking at the change, this is a trigger for a daily Submarine build, which is not done with the new pipelines. The new Submarine pipelines are just for PRs for now.

mareknovotny:
ok, so I daily builds are the old way, thanks @baldimir !

PR #458, opened 2019-05-22
Include SonarCloud analysis into OptaPlanner PR check job

rsynek:
SonarCloud analysis is about to because a part of PR checks. It already works for OptaPlanner, but so far the analysis ran in Travis, which showed to be infeasible solution due to various restrictions.

This PR enables SonarCloud analysis in Jenkins PR check jobs - so far only for OptaPlanner, but other repositories will follow.


rsynek:
Apparently, it works - see https://github.com/kiegroup/optaplanner/pull/508

PR #459, opened 2019-05-24
Sonarcloud daily jobs

rsynek:
The daily jobs for SonarCloud analysis. PR analysis cannot be done without full branch analysis, as SonarCloud might become out of sync with the repository after a couple of weeks.

Moreover, this will keep project's summary statistics up-to-date.

This initial work enables daily analysis for drools and optaplanner, other repositories will follow.

### DIFF
--- a/job-dsls/jobs/sonarcloud_daily.groovy
+++ b/job-dsls/jobs/sonarcloud_daily.groovy
@@ -1,0 +1,153 @@
+/**
+ * Creates jobs running tests with code coverage measurement enabled and reporting results together with code quality
+ * statistics into SonarCloud.
+ */
+import org.kie.jenkins.jobdsl.Constants
+
+def final DEFAULTS = [
+        ghOrgUnit              : Constants.GITHUB_ORG_UNIT,
+        branch                 : Constants.BRANCH,
+        timeoutMins            : 90,
+        ghAuthTokenId          : "kie-ci2-token",
+        label                  : "kie-rhel7 && kie-mem8g",
+        upstreamMvnArgs        : "-B -e -T1C -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true clean install",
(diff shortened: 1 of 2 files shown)


job-dsls/jobs/sonarcloud_daily.groovy
```diff
@@ -0,0 +1,153 @@
...
+ */
+import org.kie.jenkins.jobdsl.Constants
+
+def final DEFAULTS = [
+        ghOrgUnit              : Constants.GITHUB_ORG_UNIT,
+        branch                 : Constants.BRANCH,
+        timeoutMins            : 90,
+        ghAuthTokenId          : "kie-ci2-token",
+        label                  : "kie-rhel7 && kie-mem8g",
+        upstreamMvnArgs        : "-B -e -T1C -DskipTests -Dgwt.compiler.skip=true -Dgwt.skipCompilation=true -Denforcer.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true -Drevapi.skip=true clean install",
```
mareknovotny: I am not sure, but could we turn off the GWT compilation as this job is for static analysis, we don't need a real GWT compilation?

mareknovotny:
@rsynek do you want to add more kiegroup projects into sonar analysis later or are you going to add iteratively to this PR others?

job-dsls/jobs/sonarcloud_daily.groovy
rsynek: I guess that's right - and the current settings already do it, or is there any other property to be set?

rsynek:
> @rsynek do you want to add more kiegroup projects into sonar analysis later or are you going to add iteratively to this PR others?

Let's merge this one - for other repositories, I still need to provide profiles with test coverage measurement configuration. Once it's there, adding a job for every such a repository is just a matter of adding the repository on the list and providing the SonarCloud key for it.

job-dsls/jobs/sonarcloud_daily.groovy
mareknovotny: sorry, I was just mistakenly reading that option property in the opposite way ;).

PR #463, opened 2019-05-30
enabled all *.groovy scripts

mbiarnes:
nitpicking: now are enabled all *.groovy scripts in this rep - also Zanatas, dailyBuilds and releases
(we had before extra seed jobs for this - but it is not needed)

### DIFF
--- a/job-dsls/jobs/seed_job.groovy
+++ b/job-dsls/jobs/seed_job.groovy
@@ -69,7 +69,7 @@ job("${folderPath}/a-seed-job-${kieMainBranch}") {
         shell(seedJob)
 
         jobDsl {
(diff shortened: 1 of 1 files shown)


mareknovotny:
just a question what if there will be added a test job or something we would not like to have automatically created?

mbiarnes:
@mareknovotny so - we should store this groovy file in a new directory? Maybe this is possible.
Otherwise we have to maintain this list 1:1; wich each script the list gets longer.
WdyT? What is the best way here?

mbiarnes:
@mareknovotny we can create two new directories
kie-jenkins-scripts/job-dsls/jobs/jenkins
kie-jenkins-scripts/job-dsls/jobs/test
and then change the DSL scripts from job-dsls/jobs/**/*.groovy to job-dsls/jobs/jenkins/*.groovy in the seed job.
Internally when testing "./gradlew clean test" it will be tested all jobs (job-dsls/jobs/**/*.groovy) but when executing the seed job only DSL scripts in job-dsls/jobs/jenkins/*.groovy  will be executed.
The only thing is that jobs in /test should work so the testing doesn't fail, because testing is executed before a the DSL script run.

mareknovotny:
ok, i was just curious, no need to change this now

PR #466, opened 2019-05-30
corrected path where to deploy

mbiarnes:


mbiarnes:
@mareknovotny with this change it is not created in Kie/masetr another folder KIE.
The all groovy scripts was reverted as some groovy scripts are thought for another folders.
The groovy files are listed now.

PR #469, opened 2019-06-04
upgraded to next version

mbiarnes:


mbiarnes:
since this is only an upgrade of some constants I will merge it without waiting for approval

PR #479, opened 2019-06-12
job-dsls: use Maven settings with Nexus mirror

winklerm:
PR and downstream PR jobs use kieAllBuild Maven settings which specify our local Nexus mirror to speed up builds and avoid downloading artifacts from Maven Central directly.


winklerm:
@mareknovotny @mbiarnes could you please take a look?

winklerm:
Job generation succeeded, looks as expected, but I have not run the job.

PR #486, opened 2019-06-18
BXMSPROD-401: fix slave image build job DSL

winklerm:
Updated slave image build job DSL to avoid authentication issues when cloning from Gerrit.

### DIFF
--- a/job-dsls/jobs/slave_image_build.groovy
+++ b/job-dsls/jobs/slave_image_build.groovy
@@ -9,14 +9,19 @@ String command = """#!/bin/bash +x
 cd jenkins-slaves
 
 # include functionality for osbs builds
-./add-osbs.sh https://code.engineering.redhat.com/gerrit/bxms-jenkins
+# clone from gerrit moved to scm, not needed here: ./add-osbs.sh https://code.engineering.redhat.com/gerrit/bxms-jenkins
+
+rsync -av bxms-jenkins/jenkins-image-extra-bits/rhba-osbs/ansible/ ansible
+
+wget --no-check-certificate https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/userContent/packer 
+chmod u+x packer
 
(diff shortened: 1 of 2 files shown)


winklerm:
@mareknovotny @mbiarnes @AntStephenson FYI

winklerm:
Please do not merge yet, I would like to finish testing the image before (and include possible changes here).

winklerm:
I have just checked that both images have the RAW format which makes both of them 40 GB, so the size should not make a difference.

winklerm:
One final image build test job run and we will be ready for merging.

winklerm:
There are some issues with the image build job running after freshly generated from the job DSL, investigating.

job-dsls/jobs/slave_image_build.groovy
```diff
@@ -52,7 +74,7 @@ def jobDefinition = job("slave-image-build") {
             string("PSI_PASSWORD", "psi-rhba-jenkins-password")
 
             // Copies the file given in the credentials to a temporary location, then sets the variable to that location.
-            file("PSI_PRIVATE_KEY", "kie-jenkins.pem")
+            file("PSI_PRIVATE_KEY", "kie-jenkis.pem")
```
mareknovotny: typo?

winklerm:
Issues fixed by adding another provisioner for packer clearing up the trouble-making temp files. Ready to review/merge.

job-dsls/jobs/slave_image_build.groovy
winklerm: Unfortunately, this typo is in the Jenkins master configuration and I cannot rename it there (it is typo in credential ID). I would need to recreate the credential from scratch, but I need to get out the .pem file first, it cannot be easily retrieved from Jenkins UI.

winklerm:
I do not understand why I cannot remove the DO NOT MERGE label.. probably because I am missing some rights?

PR #490, opened 2019-06-27
Use no.nested.virt flavor to avoid sigsegv problem

winklerm:
Using this type of OpenStack flavor will create the instance in a special OpenStack aggregate with nested virtualization disabled, thus avoiding the situation causing SIGSEGV errors.


winklerm:
@mareknovotny @cimbalek @mbiarnes FYI

PR #494, opened 2019-07-11
Fix daily build

mbiarnes:
these PR makes only sense to merge once https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/984 was merged as it contains the modifications for the update-version-all.sh script used in kie_dailyBuild_pipeline.groovy

- all "manual" (via sed) version updates were removed since now they are done in the update script
- errai parameter were removed (and any parameter that links to it), as it is not any more needed to handle it like a daily build version - we have it right now as *.Final version


mbiarnes:
@almope @danielezonca can you check this - it is needed for the daily builds

PR #500, opened 2019-07-23
Enable sonarcloud in jBPM PR check and create a job for daily analysis of master branch.

rsynek:


job-dsls/jobs/pr_jobs.groovy
```diff
@@ -50,12 +50,6 @@ def final REPO_CONFIGS = [
         "droolsjbpm-knowledge"      : [
                 label: "kie-rhel7 && kie-mem4g"
         ],
-        "drools"                    : [
-                sonarCloudProjectKey: "org.drools:drools"
-        ],
-        "optaplanner"               : [
-                sonarCloudProjectKey: "org.optaplanner:optaplanner"
-        ],
```
MarianMacik: Just to double-check, these are leftovers, aren't they?

rsynek: Yes, as I reworked the profile for sonarcloud analysis (see kie-parent), we no longer need the project key here. Either the repository overrides the profile or it will just use `groupId:artifactId`.

rsynek: Actually this would remove the PR check jobs, I meant to remove just the sonarCloudProjectKey properties. Fixed.

rsynek:
> @rsynek looks OK but please fix the conflicts

Done

job-dsls/jobs/pr_jobs.groovy
MarianMacik: Ah, true. Thanks for explanation.

rsynek:
@mbiarnes finally no conflicts and the build check has passed. Shall we merge this?

PR #502, opened 2019-07-23
changed order and timeout for openshift-drools-hacep

mbiarnes:


mbiarnes:
makes only sense to merge after https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1004 was merged

PR #513, opened 2019-08-14
AF-2145: Introduce kie-soup to sonar with PR and daily checks

domhanak:
@rsynek @mbiarnes 
let me know if I am missing some parts


domhanak:
@rsynek I added the profile into kie-soup in https://github.com/kiegroup/kie-soup/pull/111

PR #516, opened 2019-08-20
[BXMSPROD-402] - Create seed job for cloud image builds

msmagnanijr:
See: https://issues.jboss.org/browse/BXMSPROD-402

Signed-off-by: Mauricio Magnani <mmagnani@redhat.com>


kie-ci3:
Can one of the admins verify this PR? Comment with 'ok to test' to start the build.

mareknovotny:
ok to test

mbiarnes:
@msmagnanijr could it be that there is missing an import for all the ${XXX} variables? The PR is failing because of ${PROD_COMPONENT} not found:
No such property: PROD_COMPONENT for class: script
This is the very first ${xxx} variable - I suppose all others will not be found too

msmagnanijr:
> @msmagnanijr could it be that there is missing an import for all the ${XXX} variables? The PR is failing because of ${PROD_COMPONENT} not found:
> No such property: PROD_COMPONENT for class: script
> This is the very first ${xxx} variable - I suppose all others will not be found too

hey @mbiarnes thanks for review it. As I understand it all DSLs at https://github.com/kiegroup/kie-jenkins-scripts/tree/master/job-dsls/jobs run without manual intervention, right? In this case this "Seed Job" through the parameter entered by the user, generates another Job re-using the settings of this "Seed Job". For example, if the user types "rhpam-smartrouter" (${PROD_COMPONENT} ) it will generate a job with that name and will build the "rhpam-smartrouter" image. 

In this case of "PROD_COMPONENT" I have a list of possible values.

mareknovotny:
@msmagnanijr that is right that all parameters are known in that DSL scripts. It basically can take properties from https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/src/main/groovy/org/kie/jenkins/jobdsl/Constants.groovy so that would be good to have such definition too for the cloud container images, is that reasonable ?

msmagnanijr:
> @msmagnanijr that is right that all parameters are known in that DSL scripts. It basically can take properties from https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/src/main/groovy/org/kie/jenkins/jobdsl/Constants.groovy so that would be good to have such definition too for the cloud container images, is that reasonable ?

Hello Marek, yes, and I agree with you. I added the variables for these JOBs in this last commit. I have only one question. The parameter "PROD_COMPONENT" is a list. Because there are about 9 components that must be created / executed in this JOB. In the last commit I just added one of the components (PROD_COMPONENT = "rhpam-smartrouter") to check that JOB will be created correctly but I should still work on this issue.

Have you ever had a situation like this? 

Thanks

mareknovotny:
I looked for an example and this with `MYPODS` could be good inspiration I think for your case:

```
 environment {
        MYPODS = 'pod1,pod2,pod3'
    }

    stages {
        stage("Test") {
            steps {
                script {
                    MYPODS.tokenize(',').each {
                        println "Item: ${it}"
                    }
                }
            }
        }
    }
```

Of course the iteration for every individual item in your case will create the individual job. Does it make sense to you?

mareknovotny:
Jenkins retest this please

mareknovotny:
please rebase it too

PR #517, opened 2019-08-21
Avoid deprecated Ansible yum task form

winklerm:
Replace yum task form which is now deprecated.


winklerm:
@mbiarnes @mareknovotny Please take a look, thanks!

PR #525, opened 2019-08-27
Job DSL: update UMB triggers for 7.26.x

winklerm:
Fixed the UMB trigger topic for 7.26.x.

### DIFF
--- a/job-dsls/jobs/kogito.groovy
+++ b/job-dsls/jobs/kogito.groovy
@@ -73,7 +73,7 @@ pipelineJob("$folderPath/kogito-pipeline-${mainBranch}") {
             spec ''
             providerName 'Red Hat UMB'
             overrides {
-                topic 'Consumer.rh-jenkins-ci-plugin.${JENKINS_UMB_ID}-prod-daily-master-submarine-trigger.VirtualTopic.qe.ci.ba.daily-master-submarine.trigger'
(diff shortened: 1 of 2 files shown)


winklerm:
@mbiarnes Please take a look, thanks!

job-dsls/jobs/kogito.groovy
```diff
@@ -73,7 +73,7 @@ pipelineJob("$folderPath/kogito-pipeline-${mainBranch}") {
             spec ''
             providerName 'Red Hat UMB'
             overrides {
-                topic 'Consumer.rh-jenkins-ci-plugin.${JENKINS_UMB_ID}-prod-daily-master-submarine-trigger.VirtualTopic.qe.ci.ba.daily-master-submarine.trigger'
+                topic 'Consumer.rh-jenkins-ci-plugin.${JENKINS_UMB_ID}-prod-daily-7-26-x-submarine-trigger.VirtualTopic.qe.ci.ba.daily-7-26-x-submarine.trigger'
```
mbiarnes: OK - better to remove the kogito.groovy in 7.26.x

mbiarnes: or at least not do this change - it will be orphan there as it will not be constructed via seed job

winklerm: Not sure if I get what you mean by orphan - the UMB message trigger does not require a job sending the message, it would just not be triggered.

winklerm: We need to get rid of any potential clashes in UMB topics, because it may confuse the Jenkins plugin and make mess.

PR #534, opened 2019-08-28
added branch name into kie-docs job name

mbiarnes:


job-dsls/jobs/kie_docs_pr.groovy
```diff
@@ -26,7 +26,7 @@ mavenGoals = "-B clean install"
 PrVerificationJob.addFolders(this)
 
 // Creates or updates a free style job.
-def jobDefinition = job(Constants.PULL_REQUEST_FOLDER + "/${projectName}-pullrequests")
+def jobDefinition = job(Constants.PULL_REQUEST_FOLDER + "/${projectName}" + "-pullrequests-" + Constants.BRANCH)
```
winklerm: This ``" + "`` should not be necessary to add.

winklerm: But not a blocker for merging, this branch is also already in maintenance mode, if I understand properly.

PR #536, opened 2019-08-29
removed obsolete *. groovy jobs

mbiarnes:
removed *groovy jobs that only have to exist on master branch


mbiarnes:
@rsynek DONE

PR #542, opened 2019-09-11
[KOGITO-258] - Create job to execute build and tests the Kogito image…

spolti:
…s and a PR is created

Signed-off-by: Filippe Spolti <fspolti@redhat.com>

### DIFF
--- a/job-dsls/jobs/kogito.groovy
+++ b/job-dsls/jobs/kogito.groovy
@@ -45,7 +45,12 @@ pipeline {
            steps {
                build job: "kogito-cloud-${mainBranch}", propagate: false, parameters: [[$class: 'StringParameterValue', name: 'mainBranch', value: mainBranch], [$class: 'StringParameterValue', name: 'ghOrgUnit', value: ghOrgUnit]] 
            }
-       }       
+       } 
+       stage('kogito-cloud-s2i-images') {
+           steps {
+               build job: "kogito-cloud-s2i-images-${mainBranch}", propagate: false, parameters: [[$class: 'StringParameterValue', name: 'mainBranch', value: mainBranch], [$class: 'StringParameterValue', name: 'ghOrgUnit', value: ghOrgUnit]] 
+           }
+       }         
        stage('kogito-examples') {
            steps {
                build job: "kogito-examples-${mainBranch}", parameters: [[$class: 'StringParameterValue', name: 'mainBranch', value: mainBranch], [$class: 'StringParameterValue', name: 'ghOrgUnit', value: ghOrgUnit]] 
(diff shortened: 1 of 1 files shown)


ricardozanini:
We should look for an option to externalize Jenkins reports to the outside world.

job-dsls/jobs/kogito.groovy
```diff
@@ -45,7 +45,12 @@ pipeline {
            steps {
                build job: "kogito-cloud-${mainBranch}", propagate: false, parameters: [[$class: 'StringParameterValue', name: 'mainBranch', value: mainBranch], [$class: 'StringParameterValue', name: 'ghOrgUnit', value: ghOrgUnit]] 
            }
-       }       
+       } 
+       stage('kogito-cloud-s2i-images') {
+           steps {
+               build job: "kogito-cloud-${mainBranch}", propagate: false, parameters: [[$class: 'StringParameterValue', name: 'mainBranch', value: mainBranch], [$class: 'StringParameterValue', name: 'ghOrgUnit', value: ghOrgUnit]]
```
winklerm: I guess you need to name the job differently, if I understand correctly. You plan to have 2 jobs there, right? kogito-cloud and kogito-cloud-s2i-images?

spolti: correct.

job-dsls/jobs/kogito.groovy
```diff
@@ -303,6 +308,60 @@ job("${folderPath}/kogito-cloud-${mainBranch}") {
...
+
+# clone the kogito-bom repository
+git clone https://github.com/$ghOrgUnit/kogito-cloud.git -b $mainBranch --depth 50
+# build the project
+cd s2i
+make build
+make test'''
+
+
+job("${folderPath}/kogito-cloud-${mainBranch}") {
```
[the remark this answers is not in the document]
spolti: got it.

job-dsls/jobs/kogito.groovy
```diff
@@ -303,6 +308,60 @@ job("${folderPath}/kogito-cloud-${mainBranch}") {
...
+
+
+job("${folderPath}/kogito-cloud-s2i-images-${mainBranch}") {
+    description("build project kogito-cloud-s2i")
+    parameters {
+        stringParam("mainBranch", "${mainBranch}", "Branch to clone. This will be usually set automatically by the parent trigger job. ")
+        stringParam("ghOrgUnit", "${ghOrgUnit}", "Name of organization. This will be usually set automatically by the parent trigger job. ")
+    }
+
+    label('kogito-static')
```
winklerm: Not sure if this node has all tools you need. Do you need this label specifically? Or could you use a dynamically provisioned machine (from an image which already contains osbs etc.) - for osbs, we have a label 'osbs-builder'.

job-dsls/jobs/kogito.groovy
```diff
@@ -303,6 +308,60 @@ job("${folderPath}/kogito-cloud-${mainBranch}") {
...
+                killerType 'org.jenkinsci.plugins.proccleaner.PsAllKiller'
+                killer(class: 'org.jenkinsci.plugins.proccleaner.PsAllKiller')
+                username 'jenkins'
+            }
+        }
+    }
+
+    steps {
+        environmentVariables {
+            envs(MAVEN_MIRROR_URL : "${LOCAL_NEXUS_IP}/nexus/content/groups/public")
```
winklerm:
```suggestion
            envs(MAVEN_MIRROR_URL : '${LOCAL_NEXUS_IP}/nexus/content/groups/public')
```

job-dsls/jobs/kogito.groovy
```diff
@@ -303,6 +308,60 @@ job("${folderPath}/kogito-cloud-${mainBranch}") {
 
 }
 
+
+// ++++++++++++++++++++++++++++++++++++++++++ Build and deploys kogito-cloud-s2i-images ++++++++++++++++++++++++++++++++++++++++++++++++++
+
+// definition of kogito-cloud-s2i-images script
+def kogitoCloudS2i='''#!/bin/bash -e
+
+# clone the kogito-bom repository
```
winklerm:
```suggestion
# clone the kogito-cloud repository
```

job-dsls/jobs/kogito.groovy
```diff
@@ -303,6 +308,60 @@ job("${folderPath}/kogito-cloud-${mainBranch}") {
...
+git clone https://github.com/$ghOrgUnit/kogito-cloud.git -b $mainBranch --depth 50
```
winklerm: Please consider moving the `git clone` command from shell step to a scm part of the job config - not needed to be done in this PR, it is used this way in other places as well. It would use a Jenkins plugin for cloning which is a best practice.

spolti: How and where can I do this?

winklerm: @spolti An example can be found here: https://github.com/kiegroup/kie-jenkins-scripts/blob/94368875384b24145d8681fed285f1c5586cbd6d/job-dsls/jobs/kie_dailyBuild_pipeline.groovy#L376

But I would suggest making that change in another PR, because it should affect all kogito jobs ideally, so the definitions are consistent.

spolti: it is just a copy for the kogito-cloud, shouldn't that job do it as well?

spolti: all jobs on that scripts does exactly the same thing, maybe a subsequent jira to address all other jobs will be necessary.

spolti:
seems all the tests are passing now, can it be merged so we can test the kogito-cloud images?

PR #567, opened 2019-10-01
added settings.xml for downstream and compile downstream builds

mbiarnes:


mareknovotny:
fyi @MarianMacik, @winklerm this will set settings.xml in cmd line instead of Advanced config on job configuration

PR #570, opened 2019-10-03
disabled Zanata jobs since Zanata server doesn't exist any more

mbiarnes:


almope:
Hey @mbiarnes , I would welcome any description / context / Jira to get an idea of the motivation / reason for this change.

Also, is there a plan to get back the Zanata server in the future? I would opt to remove the files for the job if the answer is no. :)

mareknovotny:
@almope there were some email thread about this demossioned Zanata server instance and we have no other replacement yet.

PR #571, opened 2019-10-03
disabled Zanata jobs since Zanata server doesn't exist any more

mbiarnes:
(cherry picked from commit 5cacd4e0e699fd6b75fe5730b8079deea121c611)


mareknovotny:
so could we get fedora zanata server setup there instead of disabling?

PR #575, opened 2019-10-08
BXMSPROD-511 Install PiG in jenkins machine

Ginxo:
BXMSPROD-511


.github/workflows/blank.yml
```diff
@@ -0,0 +1,17 @@
+name: CI
```
winklerm: This file should be probably removed?

jenkins-slaves/ansible/roles/rhba/tasks/main.yml
```diff
@@ -71,4 +71,32 @@
...
+    group: root
+
+- name: Git SSL verify to False
+  command: "git config --global http.sslVerify false"
+
+- name: Clone Product Files Generator
+  command: "git clone https://gitlab.cee.redhat.com/middleware/product-files-generator.git /opt/tools/pig/product-files-generator"
+
+- name: Package Product Files Generator
+  command: "/qa/tools/opt/maven-3.6.0-prod/bin/mvn -f /opt/tools/pig/product-files-generator/pom.xml clean install -DskipTests=true -Dmaven.repo.local=/m2/70"
```
winklerm: Is the repo /m2/70 still needed after the build? Not sure if product-files-generator.jar is a jar with dependencies or not, anyway, if the repo was not needed, it could be deleted after the build to keep the image small. But if it is required, we can keep it.

mareknovotny: no idea, @Ginxo ?

mareknovotny: it seems there were specific slave node maven repository just for product build pipeline https://issues.jboss.org/browse/BXMSPROD-59

PR #581, opened 2019-10-11
use of the same settings.xml

mbiarnes:
uses the same settings.xml now in main projects as well as upstream projects in deploy jobs (SNAPSHOTS)
7774c60d-cab3-425a-9c3b-26653e5feba1 = ci-snapshots-deploy


mareknovotny:
I am not sure we would like to change mirror from http://${env.LOCAL_NEXUS_IP}:8081/nexus/content/groups/public/ to http://${env.LOCAL_NEXUS_IP}:8081/nexus/content/groups/kie-all

mbiarnes:
@mareknovotny this was a hasty reaction/snapshot of me. I will close this one.

PR #594, opened 2019-10-23
added email function if the pipelines result was FAILURE

mbiarnes:
Adds a post action that only executes if the result of pipeline  was FAILURE.
Before, when the pipeline failed in the deploy step - the stage of sending an email was not executed since it failed before this step.
User only get an email notification when the pipeline result was SUCCESSFUL or UNSTABLE.
With this post action the user gets an email notification also when the pipelines result is FAILURE (i.e when the build&deploy fails)

### DIFF
--- a/job-dsls/jobs/DailyBuild_pipeline.groovy
+++ b/job-dsls/jobs/DailyBuild_pipeline.groovy
@@ -135,8 +135,18 @@ pipeline {
             steps {
                 cleanWs()
             }
-        }                                 
+        }                                    
     }
+    post {
+        failure{
+            script {
+                currentBuild.result = 'FAILURE\'
+            }            
+            emailext body: 'status of daily build #${BUILD_NUMBER} (${baseBranch} branch) was: ' + "${currentBuild.currentResult}" +  '\\n' +
+                    'Please look here: ${BUILD_URL} \\n' +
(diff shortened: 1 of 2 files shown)


job-dsls/jobs/DailyBuild_pipeline.groovy
```diff
@@ -135,8 +135,18 @@ pipeline {
...
-        }                                 
-    }
+        }                                    
+    }
+    post {
+        failure{
+            script {
+                currentBuild.result = 'FAILURE\'
+            }            
+            emailext body: 'status of daily build #${BUILD_NUMBER} (${baseBranch} branch) was: ' + "${currentBuild.currentResult}" +  '\\n' +
```
winklerm: Sorry for my ignorance, just curious if `currentBuild.result` or `currentBuild.currentResult` is correct? Have you tested it?

mbiarnes: sure - that is tested! On my local it worked. This is also defined in the other mailers and the bsig list received the mails. Strange to say - but currentBuild.result didn't fetch the expected value - so I did  currentResult

winklerm: Ok, cool! Thanks!

PR #607, opened 2019-10-30
fixes for community releases

mbiarnes:
1. fixed email address to bsig list
2. changed the URLs to original-repository.jboss.org since repository.jboss.org appears empty
3. removed link that was provided twice
4. added hidden variable to additional tests to assure the right URL building (see https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/script/release/kie-wbSmokeTestsMatrix.sh - the target has to be specified)

### DIFF
--- a/job-dsls/jobs/CommunityRelease_pipeline.groovy
+++ b/job-dsls/jobs/CommunityRelease_pipeline.groovy
@@ -167,28 +167,25 @@ pipeline {
                     'The artifacts are available here \\n' +
                     ' \\n' +
                     'business-central artifacts: (wildfly14.war) \\n' +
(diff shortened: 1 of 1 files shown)


mareknovotny:
ad point 2) what do you mean by empty? I thought that jbossorg team recommended to not use that origin URL.

mbiarnes:
@mareknovotny yes - but after uploading the articts o Nexus you don't see the artifacts there:
https://repository.jboss.org/nexus/content/groups/kie-group/org/kie/business-central/ look for 7.29.0.Final - and now look at https://origin-repository.jboss.org/nexus/content/groups/kie-group/org/kie/business-central/ - I am a kind of tired about this - I don't know when exactly the dir tree is updated

mareknovotny:
ok, that is then explained, just as a side effect I would report it to jbossorg team in parallel.

PR #608, opened 2019-10-31
changes for a better mail filtering

mbiarnes:
@mareknovotny changes for a better (easier) mail filtering in gmail


mareknovotny:
I guess you need to rebase, but otherwise it looks ok

PR #615, opened 2019-11-06
individual listing of groovy jobs

mbiarnes:
it is better to list jobs individual in this seed job since all here listed jobs will be created in KIE/.
Some jobs (not here listed) have to be created in another directories (i.e. Tools).
With this PR we will prevent to have jobs created twice - in different folders.


mbiarnes:
@mareknovotny never :)

PR #617, opened 2019-11-06
BXMSPROD-535: updating git repository caches

akoufoudakis:
Removing the cache repositories. Temporarily fix.

### DIFF
--- a/jenkins-slaves/ansible/roles/kie/tasks/main.yml
+++ b/jenkins-slaves/ansible/roles/kie/tasks/main.yml
@@ -145,20 +145,5 @@
   with_together:
     - "{{ firefox_long_versions }}"
     - "{{ firefox_short_versions }}"
-
-### Git repositories ###
-
-- name: Create git-repos dir
-  file:
-    path: /home/jenkins/git-repos
-    state: directory
-
-- name: Symlink jbpm-console-ng -> jbpm-wb
-  file:
-    path: /home/jenkins/git-repos/jbpm-console-ng.git
-    src: /home/jenkins/git-repos/jbpm-wb.git
-    state: link
-    owner: jenkins
-    group: jenkins
-    mode: 0755
+    
 ...
--- a/jenkins-slaves/ansible/roles/kie/tasks/main.yml
+++ b/jenkins-slaves/ansible/roles/kie/tasks/main.yml
@@ -157,6 +157,7 @@
   with_items:
    - kiegroup/lienzo-core
    - kiegroup/lienzo-tests
+   - errata/errata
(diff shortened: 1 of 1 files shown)


jenkins-slaves/ansible/roles/kie/tasks/main.yml
```diff
@@ -153,49 +153,6 @@
     path: /home/jenkins/git-repos
```
almope: Do we need still the directory to be created?

akoufoudakis: Oh. It slipped through. I removed the Jenkins permissions for git-repos, but missed that one. Correcting right now.

mareknovotny:
there is some misunderstanding here with what the jira requests
we should not remove git cache repos, but update them to be in sync with repository-list.txt file list

akoufoudakis:
As discussed, repos need to be updated, but not the removed.

jenkins-slaves/ansible/roles/kie/tasks/main.yml
```diff
@@ -158,34 +157,30 @@
   with_items:
    - kiegroup/lienzo-core
    - kiegroup/lienzo-tests
-   - errai/errai
```
mareknovotny: this one I would keep until we remove Errai from Upstream builders

jenkins-slaves/ansible/roles/kie/tasks/main.yml
```diff
@@ -158,34 +157,30 @@
...
+   - kiegroup/droolsjbpm-build-bootstrap
    - kiegroup/kie-soup
    - kiegroup/appformer
-   - kiegroup/droolsjbpm-build-bootstrap
    - kiegroup/droolsjbpm-knowledge
    - kiegroup/drools
    - kiegroup/optaplanner
-   - kiegroup/optaweb-employee-rostering
    - kiegroup/jbpm
-   - kiegroup/jbpm-work-items
```
mareknovotny: this should not be removed IMHO, see https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/script/repository-list.txt#L20

jenkins-slaves/ansible/roles/kie/tasks/main.yml
```diff
@@ -204,4 +199,5 @@
     owner: jenkins
     group: jenkins
     mode: 0755
-...
\ No newline at end of file
+...
+...
```
mareknovotny: doubled 3 dots? Is this intentional?

PR #626, opened 2019-11-08
added more mem for droolsjbpm-integration PR

mbiarnes:


mareknovotny:
Even adding more memory without proper analysis where it was coming is not systematic ;)

PR #641, opened 2019-11-27
2nd mail if a re-build were done after sanity checks

mbiarnes:
in case the pipeline was aborted and a new build required (i.e. after several days of sanity checks) a new mail will be send out, but this time not again the sanity test one.


mbiarnes:
do not yet merge - still doing some tests

PR #642, opened 2019-11-27
requested global git configuration

mbiarnes:
when trying to push from an arbitrary node something to gerrit a 
git config --global is required


mbiarnes:
do not yet merge - still doing some tests

PR #648, opened 2019-12-09
Community pipeline

mbiarnes:
... the build itself is not dependant if the branch already exists or not - but if the repBuild variable is selected and =YES 
pipeline crashed because branch was not existing - so repBuild=0 - the deploy stage was never activated
the constants have to be modified since the version on master is 7.32.0-SNAPSHOT now


mbiarnes:
Jenkins please retest this

PR #650, opened 2019-12-10
renamed groovy script

mbiarnes:


mareknovotny:
I am ok to rename the script, but please use more meaningful name instead of rhdm. 
My suggestions: `prod_seed_cloud_jobs.groovy` or `seed_nighly_cloud_jobs.groovy` or `seed_prodnightly_cloud.groovy` and let know original author too ;)

PR #665, opened 2020-01-14
changed mail recipients

mbiarnes:
only last mail after sanity checks - when the release was closed and done will be send to bsig list


mbiarnes:
@mareknovotny have to add some changes there - this PR is not ready yet for merging

PR #667, opened 2020-01-15
Make trigger phrases for downstream builds more benevolent

rsynek:
Regex checked via https://regex101.com/


mareknovotny:
have you tried on real job? you can add this regex there  ;), I do not know if Jenkins regex is pure Java or not, we can try it ;)

mbiarnes:
@rsynek LGTM. I expect you tested it somehow and you are sure that this reg expr. works in these groovy files? At least ./gradlew clean test run properly without any failures.

rsynek:
Not yet tested in a Jenkins job; not sure if `./gradlew clean test` can reveal any issue. If the regex is not correct, it just won't trigger any job.

@mareknovotny @mbiarnes can you think of a proper way of testing this? I have just this idea: setting up a dummy job, which reacts on this PR and typing the command here in a comment.

mbiarnes:
@rsynek will do this

mbiarnes:
@rsynek tested and it works

PR #676, opened 2020-01-22
created dir for binaries to upload to filemgmt

mbiarnes:
- added comments for each step
- added a new stage where a directory (${kieVersion}"_uploadBinaries) is created where the binaries for uploading to filemgmt.jboss.org are stored - later this directory is compressed and archieved)
- a flat file with the content and size of the new dir is created and stored into it - so it is available in the *.tar.gz too
- a new hidden parameter deployPath is created
- due to the big sized file (*.tar.gz, +/- 2G) that is archived the log rotator stores only 3 builds.

### DIFF
--- a/job-dsls/jobs/communityRelease_pipeline.groovy
+++ b/job-dsls/jobs/communityRelease_pipeline.groovy
@@ -49,6 +49,7 @@ pipeline {
                 }    
             }
         }
+        // checks if release branch already exists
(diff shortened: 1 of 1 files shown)


mbiarnes:
merge together: 
https://github.com/kiegroup/kie-jenkins-scripts/pull/676
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1168

PR #683, opened 2020-01-28
disabled 7.30.x jobs

mbiarnes:
- all 7.30.x PRs disabled
- seed job for 7.30.x disabled
- 7.30.x daily builds are not any more triggered
- all other jobs haven't a automatic trigger and have to be run manually


mbiarnes:
this PR will be closed since folders 6.5, 7.32.x and 7.30.x will be removed on Jenkins directly - so there are no any more seed jobs

PR #691, opened 2020-02-13
attached build log for PRs

mbiarnes:
@mareknovotny @ge0ffrey the mail send for a failed or unstable PR will have its log attached in a zipped format. Not any more as text in its mail.


ge0ffrey:
Good step forward, fixes the catch 22 feeling.

PR #712, opened 2020-03-19
Permissions

mbiarnes:


mbiarnes:
@cristianonicolai Started a job with this new machine with Maven upgrade and used  3.6.2 and it failed not because of the Maven version, becasue another reason. Also the kogito-deploy is failing.
So nothing to do with the Maven upgrade. There is already a machine with this Maven update. The merge of this PR is for upgrading our jenkins-slaves. Next time someone runs the jenkins-slave job the upgrades will be fetched.

PR #714, opened 2020-03-20
Remove synchronized-resolver Maven ext for recent Maven versions

winklerm:
The synchronized-resolver [1] Maven extension causes NPE with Maven >= 3.6.2. The extension claims to resolve the bug [2] but according to comments at [2], the original problem has not been observed on recent Maven versions, therefore we decided to discontinue using this extension for Maven >= 3.6.2.

Also adds Maven 3.6.3 because of a few regression fixes.

[1] https://github.com/psiroky/maven-parallel-resolver
[2] https://issues.apache.org/jira/browse/MNG-5315


winklerm:
@mareknovotny FYI

jenkins-slaves/ansible/kie-rhel7.yml
```diff
@@ -4,7 +4,8 @@
   gather_facts: false
 
   vars :
-    maven_versions: [3.3.9,3.5.0,3.5.2,3.5.4,3.6.0,3.6.2]
+    maven_versions: [3.3.9,3.5.0,3.5.2,3.5.4,3.6.0,3.6.2,3.6.3]
+    maven_versions_needing_extension: [3.3.9,3.5.0,3.5.2,3.5.4,3.6.0]
```
mareknovotny: there is mentioned in the https://issues.apache.org/jira/browse/MNG-5315 that it should work without that parallel solver Maven from version 3.5.0 we could reduce the existing list to that 3.3.9 only ;)

PR #716, opened 2020-03-23
removed superfluous scripts

mbiarnes:
@cristianonicolai since there is now the Jenkinsfile.deploy in kogito-runtimes that does the deploy of the wanted kogito reps this groovy script isn't needed any more

@Ginxo I saw both pigExecution jobs in https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Tools/ were executed the last time in October or November of last year. So I expect we don't need them any more? If so, we should remove also the groovy scripts and change the a-seed-job - and take out this groovy job.


Ginxo:
@mbiarnes I would keep pigExecution groovy script because we are planning to use it in the future to automatise some RHBA productization process steps.

mbiarnes:
@Ginxo OK

PR #718, opened 2020-03-23
mails will not be sent every time when build succeeds - only when prev…

mbiarnes:
…ious build status was FAILURE or  UNSTABLE


mbiarnes:
Jenkins please retest this

PR #719, opened 2020-03-24
added stage for sending an email in case the deploy fails

mbiarnes:
@mareknovotny I saw in the last execution of community release that an email is not send if the build fails. I added this stage here.

### DIFF
--- a/job-dsls/jobs/communityRelease_pipeline.groovy
+++ b/job-dsls/jobs/communityRelease_pipeline.groovy
@@ -144,6 +144,18 @@ pipeline {
                 }
             }
         }
+        stage ('Send mail only if build fails') {
+            when{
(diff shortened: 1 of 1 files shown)


mareknovotny:
won't the email be sent then twice? I see there are 1st and 2nd email kept there

mbiarnes:
@mareknovotny tricky:
1st email is send if the build was SUCCESSFUL/UNSTABLE
2nd mail is send only when the pipeline ws aborted and re-triggered and the build is SUCCESSFUL/UNSTABLE
in case the build fails the pipeline doesnt reach these steps and no mail is sent - this PR adds a email function only in case the build fails

mbiarnes:
@mareknovotny please let me merge this since I am doing still some test with #720 and I don't want the kie-jenkins-scripts PR overwritten.

PR #721, opened 2020-03-30
this change throws an exception - after test it will be removed

mbiarnes:
Please do not comment or merge this PR. It is for test purposed only and will be closed when the tests were successful


mbiarnes:
Jenkins please retest this

PR #729, opened 2020-04-08
BXMSPROD-710 new File(String) allowed

Ginxo:
BXMSPROD-710


Ginxo:
not needed since fileExists can be used

PR #731, opened 2020-04-24
BXMSPROD-533-FINISHING

akoufoudakis:
Last finishing additions to the job.

### DIFF
--- a/job-dsls/jobs/srcclr_scan_job.groovy
+++ b/job-dsls/jobs/srcclr_scan_job.groovy
@@ -2,13 +2,13 @@ import org.kie.jenkins.jobdsl.Constants
 
 def repoList = new URL(Constants.REPO_FILE_URL).text.readLines()
 
-def srcclrPath = Constants.SRCCLR_JOBS_FOLDER
+def srcclrPath =
 
-folder("${srcclrPath}")
+folder(Constants.SRCCLR_JOBS_FOLDER)
 
 for (repo in repoList) {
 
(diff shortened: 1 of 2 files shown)


job-dsls/jobs/srcclr_scan_job.groovy
```diff
@@ -8,7 +8,7 @@ folder("${srcclrPath}")
 
 for (repo in repoList) {
 
-    def jobName = "srcclr-scan-${repo}"
+    def jobName = "${srcclrPath}/srcclr-scan-${repo}"
```
winklerm, line 11: Just a very minor comment, maybe the constant _SRCCLR_JOBS_FOLDER_ could be used here directly and the variable _srcclrPath_ is not needed, as it is in [1]?

[1] https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/jobs/pr_jobs.groovy#L161

job-dsls/jobs/srcclr_scan_pipeline.groovy
```diff
@@ -45,8 +45,8 @@ parallel branches
         '''
 
 folder("${srcclrPath}")
-
-pipelineJob('srcclrpipeline') {
+def pipelineName = "${srcclrPath}/srcclrpipeline"
+pipelineJob(pipelineName) {
```
Ginxo, line 49: why don't you change this by `pipelineJob("${srcclrPath}/srcclrpipeline")` and remove the `pipelineName` variable.

PR #735, opened 2020-04-30
upgraded Gradle version and Jenkins version and plugins

mbiarnes:


mbiarnes:
please merge this PR 
AFTER #736 
otherwise it will create all FDB again without being disabled

PR #738, opened 2020-05-11
upgraded constants to 7.38.0

mbiarnes:


job-dsls/src/main/groovy/org/kie/jenkins/jobdsl/Constants.groovy
```diff
@@ -24,7 +24,7 @@ class Constants {
     static final String GITHUB_ORG_UNIT = "kiegroup"
     static final String PULL_REQUEST_FOLDER = "pullrequest"
     static final String DEPLOY_FOLDER = "deployedRep"
-    static final String KIE_PREFIX = "7.36.0"
+    static final String KIE_PREFIX = "7.38.0"
```
mareknovotny, line 27: hmm, we skipped 7.37.0 ;)

mbiarnes: yes - we or I simply forget

PR #740, opened 2020-05-11
disabled PRs and CDBs and removed groovy files from seed-job

mbiarnes:


mbiarnes:
when this is merged I will enable the see-job again - right now it id disabled to prevent PRs and CDBs overwritings

mbiarnes:
the PRs are enabled again since there are changes referring the github user - we need this change in the PRs

PR #748, opened 2020-05-21
disabled obsolete pr-jobs and fixed seed job not to create them any more

mbiarnes:


mbiarnes:
the this is part of #754 - so this PR can be closed

PR #750, opened 2020-05-28
added missing $ to kieVersion

mbiarnes:


mbiarnes:
link in mail was not created correctly

PR #753, opened 2020-06-16
 Changed roles to accomodate the changes for the kogito-cloud pipelines

Kaitou786:
Closses [KOGITO-2383] and [KOGITO-2384]
see: https://issues.redhat.com/browse/KOGITO-2383 && \
https://issues.redhat.com/browse/KOGITO-2384

Signed-off-by: Tarun Khandelwal <tarkhand@redhat.com>

### DIFF
--- a/jenkins-slaves/ansible/data/config/bashrc
+++ b/jenkins-slaves/ansible/data/config/bashrc
@@ -7,7 +7,9 @@ fi
 
 export LANG=en_US.UTF-8
 export JAVA_HOME=/opt/tools/jdk1.8
-export MAVEN_HOME=/opt/tools/apache-maven-3.2.5
+export MAVEN_HOME=/opt/tools/apache-maven-3.5.2
 export MAVEN_OPTS="-Xmx2g"
-
-export PATH=$JAVA_HOME/bin:$MAVEN_HOME/bin:$PATH
+export GRAALVM_HOME=/opt/tools/graalvm-ce-java11-19.3.1
+export GOROOT=/opt/tools/go
+export GOPATH=/home/jenkins/go
+export PATH=$JAVA_HOME/bin:$MAVEN_HOME/bin:$GOROOT/bin:$GRAALVM_HOME/bin:$GOPATH/bin:/opt/tools/mercurial-5.3.1/:/opt/tools/extras/:$PATH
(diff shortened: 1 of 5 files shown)


jenkins-slaves/ansible/data/config/bashrc
```diff
@@ -7,7 +7,9 @@ fi
 
 export LANG=en_US.UTF-8
 export JAVA_HOME=/opt/tools/jdk1.8
-export MAVEN_HOME=/opt/tools/apache-maven-3.2.5
+export MAVEN_HOME=/opt/tools/apache-maven-3.5.2
```
winklerm, line 10: I saw somewhere we started to use Maven 3.6.3(?), would it make sense to upgrade to it here as well?

jenkins-slaves/ansible/roles/common/tasks/main.yml
```diff
@@ -127,6 +135,102 @@
   with_items:
     - "{{ maven_versions }}"
 
+### GraalVM ###
+
+- name: Dowload GraalVM archives
+  get_url:
+    url: https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-19.3.1/graalvm-ce-java11-linux-amd64-19.3.1.tar.gz
```
winklerm, line 142: It would be great to extract the versions into constants to avoid repeating them in download, extract, etc. steps.

jenkins-slaves/ansible/roles/kogito/tasks/main.yml
```diff
@@ -17,4 +17,34 @@
 
 - name: Install Yarn
   shell: npm install -g yarn --registry=${NPM_REGISTRY_URL}
+
+### Configure jenkins user to use podman ###
+- name: Copy podman cofig file
```
winklerm, line 22:
```suggestion
- name: Copy podman config file
```

jenkins-slaves/ansible/data/config/bashrc
Kaitou786: Ah yes, I think so. I'll update this to 3.6.3

mareknovotny: kogito aka quarkus requires at least Maven 3.6.2, so good to use the latest what we have available

spolti: we should use at least 3.6.0

Kaitou786: Yes, updated the file to use latest available version(3.6.3) for maven

PR #754, opened 2020-06-18
Tiny fixes

mbiarnes:
with the removal of pr_jobs.groovy and the upgrade of Maven version we are sure tha the jobs are not overridden again and the "old" Maven 3.5.2 version applied.
As FDBs, downstream.production and upstream builds are using [Jenkinsfile](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/Jenkinsfile#L16) they use Maven 3.5.4 (IIRC this was a buggy version of Maven). Not clear to me where this [constant](https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/src/main/groovy/org/kie/jenkins/jobdsl/Constants.groovy#L21) is used. So I left it like it is. To have in mind that we are using somehow right now three different maven version.
I would syncr. all Mavens to 3.6.3.

### DIFF
--- a/job-dsls/jobs/communityRelease_pipeline.groovy
+++ b/job-dsls/jobs/communityRelease_pipeline.groovy
@@ -2,13 +2,13 @@ import org.kie.jenkins.jobdsl.Constants
 
 def kieVersion=Constants.KIE_PREFIX
 def baseBranch=Constants.BRANCH
-def releaseBranch="r7.29.0.Final"
+def releaseBranch="r7.40.0.Final"
 def organization=Constants.GITHUB_ORG_UNIT
 def m2Dir = Constants.LOCAL_MVN_REP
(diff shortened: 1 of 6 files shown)


mareknovotny:
can you please also announce this change for our ML ?

mareknovotny:
it would be also very nice to upgrade jdk 8 to the latest one, but I think it is ok to do it in different PR as this needs to create new image for build nodes too.

mbiarnes:
@mareknovotny rebased!
The seed-job on Jenkins execute the pr_job.groovy and compile_downstream_build.groovy. 
Once this seed-job is executed it also "updates" seed_job-groovy - so the seed-job itself. It will be overwritten. This has not any more the jobs  pr_job and compile_downstream!

PR #762, opened 2020-06-30
reduced amount of executions to store

mbiarnes:


job-dsls/jobs/errai_pr.groovy
```diff
@@ -31,7 +31,7 @@ job(jobName) {
     description("Runs CI build against PRs submitted to the Errai repository (github.com/errai/errai.git)")
 
     logRotator {
-        daysToKeep(7)
+        daysToKeep(3)
```
mareknovotny, line 34: maybe better to have last builds not the days as it is rare to have PR here

mbiarnes: @mareknovotny done

PR #764, opened 2020-06-30
Last5executions 7.39.x

mbiarnes:


job-dsls/src/main/groovy/org/kie/jenkins/jobdsl/Constants.groovy
```diff
@@ -17,8 +17,8 @@ package org.kie.jenkins.jobdsl
 
 
 class Constants {
-    static final String MAVEN_VERSION = "3.6.3"
-    static final String UPSTREAM_BUILD_MAVEN_VERSION = "3.6.3"
+    static final String MAVEN_VERSION = "3.5.2"
```
winklerm, line 20: Is downgrading Maven intentional?

mareknovotny: I think so, product on 7.39.x is not read to migrate, PNC does use 3.5.2 too, we can change that for 7.9.0

PR #768, opened 2020-07-03
remove additional artifacts

Ginxo:
`archiveArtifacts` removed because they are already defined here https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/Jenkinsfile#L102 and her https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/Jenkinsfile#L105
related with https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1383


mbiarnes:
@Ginxo so I understand that additionalArtifacts will only have it's "values" in the new_pr_job.groovy. All other new jobs wwill have the value additionalArtifacts [ ] - so empty.

Ginxo:
@mbiarnes indeed because they are already defined in Jenkinsfile and the need to be splitted in several archiveArtifacts executions

PR #781, opened 2020-07-09
since a FDB for droolsjbpm-tools was requested it was added

mbiarnes:
A FDB for droolsjbpm-tolls was requested - we simply hadn't one job for this.
Now it will be generated. As @jomarko told, this rep is rarely touched.
But to have it right now doesn't hurt. I the worst case this job never is started.


mareknovotny:
we don't need this repository in FDB, it was already added to CDB and it should be enough or not?

mareknovotny:
see this PR we had merged https://github.com/kiegroup/kie-jenkins-scripts/pull/757 for CDB

mbiarnes:
added droolsjbpm-tools as downstream for FDBs and CDBs

PR #783, opened 2020-07-13
Include RH CAs into kie-rhel* image

winklerm:
We need to include the needed CAs into the kie-rhel* image in order to be able to use https to specific servers.


winklerm:
FYI @mareknovotny @mbiarnes

winklerm:
It seems this actually does not solve the original issue, still investigating.

PR #790, opened 2020-07-17
shortened passphrases to reduce typo errors

mbiarnes:
People start to complain that the pass phrase are very long and that the "please" is superfluous.
I think of shorten the pass phrases.
**IMPORTANT**: before merging thiswe have to send a mail to BSIG with this new pass phrases.

### DIFF
--- a/job-dsls/jobs/new_compile_downstream_build.groovy
+++ b/job-dsls/jobs/new_compile_downstream_build.groovy
@@ -127,7 +127,7 @@ for (repoConfig in REPO_CONFIGS) {
                 orgslist("${ghOrgUnit}")
                 whitelist("")
                 cron("")
-                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) compile downstream build.*")
+                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) compile db.*")
                 allowMembersOfWhitelistedOrgsAsAdmin(true)
                 whiteListTargetBranches {
                     ghprbBranch {
--- a/job-dsls/jobs/new_downstream_pr_jobs.groovy
+++ b/job-dsls/jobs/new_downstream_pr_jobs.groovy
@@ -124,7 +124,7 @@ for (repoConfig in REPO_CONFIGS) {
                 orgslist("${ghOrgUnit}")
                 whitelist("")
                 cron("")
-                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) full downstream build.*")
+                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) full db.*")
                 allowMembersOfWhitelistedOrgsAsAdmin(true)
                 whiteListTargetBranches {
                     ghprbBranch {
--- a/job-dsls/jobs/new_downstream_production.groovy
+++ b/job-dsls/jobs/new_downstream_production.groovy
@@ -120,7 +120,7 @@ for (repoConfig in REPO_CONFIGS) {
                 orgslist("${ghOrgUnit}")
                 whitelist("")
                 cron("")
-                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) full downstream production build.*")
+                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) prod downstream.*")
                 allowMembersOfWhitelistedOrgsAsAdmin(true)
                 whiteListTargetBranches {
                     ghprbBranch {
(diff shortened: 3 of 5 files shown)


Ginxo:
I more or less agree with the `please`, but I don't see the point to short the rest

mareknovotny:
I don'ŧ insist on please, it can be just optional so any polite person can add that, but it should not be requested there

job-dsls/jobs/new_compile_downstream_build.groovy
```diff
@@ -127,7 +127,7 @@ for (repoConfig in REPO_CONFIGS) {
                 orgslist("${ghOrgUnit}")
                 whitelist("")
                 cron("")
-                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) compile downstream build.*")
+                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) compile db.*")
```
mareknovotny, line 130: if we would like to get shorter one then I suggest:
```suggestion
                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) cdb.*")
```
or
```suggestion
                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) full compile.*")
```

job-dsls/jobs/new_downstream_pr_jobs.groovy
```diff
@@ -124,7 +124,7 @@ for (repoConfig in REPO_CONFIGS) {
                 orgslist("${ghOrgUnit}")
                 whitelist("")
                 cron("")
-                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) full downstream build.*")
+                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) full db.*")
```
mareknovotny, line 127: if we would like to get shorter one then I suggest:
```suggestion
                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) fdb.*")
```

job-dsls/jobs/new_downstream_production.groovy
```diff
@@ -120,7 +120,7 @@ for (repoConfig in REPO_CONFIGS) {
                 orgslist("${ghOrgUnit}")
                 whitelist("")
                 cron("")
-                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) full downstream production build.*")
+                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) prod downstream.*")
```
mareknovotny, line 123: if we would like to get shorter one then I suggest:
```suggestion
                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) full prod build.*")
```
or
```suggestion
                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) product FDB.*")
```

job-dsls/jobs/new_pr_jobs.groovy
```diff
@@ -201,7 +201,7 @@ for (repoConfig in REPO_CONFIGS) {
                 orgslist("${ghOrgUnit}")
                 whitelist("")
                 cron("")
-                triggerPhrase(".*[j|J]enkins,?.*(retest|test) this please.*")
+                triggerPhrase(".*[j|J]enkins,?.*(retest|test) this.*")
```
mareknovotny, line 204: this is ok to remove please word even to be polite you can still include it no matter what phrase reges is ;)

job-dsls/jobs/new_upstream.groovy
```diff
@@ -120,7 +120,7 @@ for (repoConfig in REPO_CONFIGS) {
                 orgslist("${ghOrgUnit}")
                 whitelist("")
                 cron("")
-                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) upstream build.*")
+                triggerPhrase(".*[j|J]enkins,?.*(execute|run|trigger|start|do) upstream.*")
```
mareknovotny, line 123: thinking why we have this to be triggered by comment phrase?
I don't see a reason for it or please explain it to me ;)

job-dsls/jobs/new_downstream_production.groovy
AntStephenson: I guess option B but lowercase to be consistent with the others.

AntStephenson:
I quite like the short hand suggests by @mareknovotny

Ginxo:
Well from my point of view there's no reason to change "launch" expressions, I think it will generate problems for the ones already use to them. In case we want them in any case I would suggest changes like

- `.*[j|J]enkins,?.*(execute|run|trigger|start|do) (prod|product) (downstream|FDB).*`
- `.*[j|J]enkins,?.*(execute|run|trigger|start|do) (full downstream build|full db@FDB).*`
and so on... So it's just to support old ones and shorter options.

I wouldn't go for `please` removing since the addition to default trigger phrase solved the problem of triggering everything

mbiarnes:
close because the PR #803 does this

PR #791, opened 2020-07-17
shortened passphrases to reduce typo errors

mbiarnes:
People start to complain that the pass phrase are very long and that the "please" is superfluous.
I think of shorten the pass phrases.
**IMPORTANT**: before merging thiswe have to send a mail to BSIG with this new pass phrases.


mbiarnes:
closed because the PR #803 does this

PR #796, opened 2020-07-29
fixed wrong branch for deployment of Optaplanner snapshots

mareknovotny:
https://issues.redhat.com/browse/BXMSPROD-888


mareknovotny:
@Ginxo i looked through all jobs, which are not on new layout done by @mbiarnes, but there is not such thing like here there the old DSL definition is used, but for sure if we discover any other issue we need to fix it too. Right now this is more like hotfix, so merging ASAP is required as half of deploy jobs relying on Optaplanner snapshots are broken

PR #798, opened 2020-08-03
BXMSPROD-888: updated deploy job for optaplanner to use right branch

mbiarnes:
these changes create all deploy jobs ho we want.
- drools triggers optaplanner-7.x
- optaplanner-7.x has as upstream branch master and current 7.x
- the triggered jobs of optaplanner-7.x are fine


mbiarnes:
also tested drools and jbpm, the jobs before and after optaplanner if they are still OK with master for upstream and master for current

PR #808, opened 2020-08-31
BXMSPROD-984: renamed jobs

mbiarnes:


mbiarnes:
Related PR https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1449

mbiarnes:
When this gets merged we have to do on Jenkins at the same time:
1. disable all existing jobs - so no job with "new" starts anymore
2. move all running jobs in pullrequest, fullDownstream, compileDownstream, downstream-production and upstream to pullrequestBack -  so no information about running jobs gets lost
3. remove everything what is in the folders pullrequest, fullDownstream, compileDownstream, downstream-production and upstream
4. change online the seed job so the jobs without "new" are taken also the first time it executes

mareknovotny:
> When this gets merged we have to do on Jenkins at the same time:
> 
>     1. disable all existing jobs - so no job with "new" starts anymore
> 
>     2. move all running jobs in pullrequest, fullDownstream, compileDownstream, downstream-production and upstream to pullrequestBack -  so no information about running jobs gets lost

We already have there backup from the previous kie-jenkins-scripts definitions or not? this would overwrite them or am i wrong?
> 
>     3. remove everything what is in the folders pullrequest, fullDownstream, compileDownstream, downstream-production and upstream
> 
>     4. change online the seed job so the jobs without "new" are taken also the first time it executes

mbiarnes:
@mareknovotny yes - but the jobs in there are disabled and a few weeks old - I would say we don't need any more the jobs there - wdyt?

mareknovotny:
@mbiarnes I think our intention was to keep backups from the original ones, which were fully defined in DSL not in pipeline to have a comparison to check. Everything else is in git history ;)
So from my POV we don't need to backup the current jobs after renaming. But do what you think is good for future restoration

PR #831, opened 2020-10-14
Fixed product version to not conflict with 7.44.x branch prod nightly

mareknovotny:


job-dsls/src/main/groovy/org/kie/jenkins/jobdsl/Constants.groovy
```diff
@@ -38,7 +38,7 @@ class Constants {
     static final String SONARCLOUD_FOLDER = "sonarcloud"
     static final String REPORT_BRANCH = "7.x"
```
mareknovotny, line 39: not sure if this needs to be changed too? @mbiarnes

winklerm (approved):
Not sure about this particular value, but it makes sense.

PR #867, opened 2020-11-27
Use JDK 11 for optawebs and optaplanner-quickstarts

Christopher-Chianelli:
All of the opta repo's use release=8, but "release" is only available in JDK 11+ (not in JDK 8). So they need to be build with JDK 11, even though their output are JDK 8 compatible.

### DIFF
--- a/job-dsls/jobs/deploy_jobs.groovy
+++ b/job-dsls/jobs/deploy_jobs.groovy
@@ -194,7 +194,10 @@ for (repoConfig in REPO_CONFIGS) {
             }
         }
 
-        if (repo == "optaplanner") {
+        if (repo == "optaplanner" ||
+            repo == "optaweb-employee-rostering" ||
+            repo == "optaweb-vehicle-routing" ||
+            repo == "optaplanner-quickstarts") {
             jdk("kie-jdk11")
         } else {
             jdk("kie-jdk1.8")}
--- a/job-dsls/jobs/deploy_jobs.groovy
+++ b/job-dsls/jobs/deploy_jobs.groovy
@@ -152,6 +152,7 @@ for (repoConfig in REPO_CONFIGS) {
     String ghOrgUnit = get("ghOrgUnit")
     String ghAuthKey = get("ghAuthKey")
     String zulipStream = get("zulipNotificationStream")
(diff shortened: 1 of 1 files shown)


ge0ffrey (approved):

job-dsls/jobs/deploy_jobs.groovy
```diff
@@ -194,7 +194,10 @@ for (repoConfig in REPO_CONFIGS) {
             }
         }
 
-        if (repo == "optaplanner") {
+        if (repo == "optaplanner" ||
+            repo == "optaweb-employee-rostering" ||
+            repo == "optaweb-vehicle-routing" ||
+            repo == "optaplanner-quickstarts") {
```
ge0ffrey, line 200: All of these use release=8, but "release" is only available in JDK 11+ (not in JDK 8). So they need to be build with JDK 11, even though they are JDK 8 compatible.

job-dsls/jobs/deploy_jobs.groovy
```diff
@@ -194,7 +194,10 @@ for (repoConfig in REPO_CONFIGS) {
             }
         }
 
-        if (repo == "optaplanner") {
+        if (repo == "optaplanner" ||
```
Ginxo, line 197: I propose you to change it by
`if (["optaplanner", "optaweb-employee-rostering", "optaweb-vehicle-routing", "optaplanner-quickstarts"].contains(repo))`

Ginxo (approved):
LGTM, just check my proposal

mbiarnes (approved):
+1

job-dsls/jobs/deploy_jobs.groovy
mareknovotny: or we can define array variable and do the calling method on it as it would be more readable and clean IMHO

something like
```
def OPTA_REPOS = ["optaplanner", "optaweb-employee-rostering", "optaweb-vehicle-routing", "optaplanner-quickstarts"];
if (OPTA_REPOS.contains(repo)) ...
```

mareknovotny (commented):
I think there could be more readable form of the if condition

Christopher-Chianelli:
Done (used `String[] jdk11Repos = ["optaplanner", "optaweb-employee-rostering", "optaweb-vehicle-routing", "optaplanner-quickstarts"]`)

mareknovotny:
jenkins retest this

mareknovotny (approved):

PR #868, opened 2020-11-30
upgraded master to 7.48.0 prefix

mbiarnes:
The prefix of our artifacts we create in daily builds should be **7.48.0** since master branch was updated to 7.48.0-SNAPSHOT
The nightly version has **7.47.0.20201130-051112**. It still uses the prefix **7.47.0** - this have to be changed.


mareknovotny:
@mbiarnes please use template as a help for not forget on required information. Otherwise remove templated PR comment. But I would like to see always a good description for quick and good review input

mareknovotny (approved):

PR #902, opened 2021-01-11
disabling jdk11 pipeline job for not ready branch

mareknovotny:
The jdk 11 pipeline job now makes sense only for master branch so disabling on 7.48.x where we know it won't work.

<pre>

</pre>

### DIFF
--- a/job-dsls/jobs/dailyBuild_jdk11_pipeline.groovy
+++ b/job-dsls/jobs/dailyBuild_jdk11_pipeline.groovy
@@ -179,8 +179,9 @@ pipeline {
 
 pipelineJob("${folderPath}/daily-build-jdk11-pipeline-${baseBranch}") {
 
-    description('this is a pipeline job for the daily build of all reps')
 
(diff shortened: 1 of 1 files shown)


mbiarnes (approved):
+1

mbiarnes:
@mareknovotny to disable complete the jdk11 pipeline on 7.48.x we also have to do this on https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/jobs/kieAll_meta_pipeline.groovy#L39 {where a errro is since it has to state 7.48.x instead of master} But this line can be removed easily.
Attention. Rght now we know that the daily builds are triggered bu a cron job since the UMB messaged don't work at all. I will meanwhile change this cron job so the jdk 11 pipeline will not run for 7.48.x.

mareknovotny:
hmm, it seems to me chaotic to have metadata UMB trigerring different branch jobs :/
I think there should be used that metadata job for every branch, WDYT?

PR #910, opened 2021-01-18
Community binaries upload

mbiarnes:

**JIRA**: _(please edit the JIRA link if it exists)_ 

[BXMSPROD-1162](https://issues.redhat.com/browse/BXMSPROD-1162)

Since the step that uploads the binaries to Nexus fails in 70% of all executions because of network failures (when it is to slow the upload takes to long and we get an 401 error) the community pipeline fails and the last steps have to be done manually.
We use now the same approach as on daily builds - zip a file, upload the zipped file that is de-compressed automatically on Nexus.
This was tested bu a fake release and the step succeeded.

<pre>

</pre>

### DIFF
--- a/job-dsls/jobs/communityRelease_pipeline.groovy
+++ b/job-dsls/jobs/communityRelease_pipeline.groovy
@@ -27,6 +27,7 @@ folder("community-release")
 def folderPath="community-release"
 
 def comRelease='''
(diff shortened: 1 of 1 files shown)


mareknovotny (approved):
lgtm

PR #930, opened 2021-02-11
PROD jobs and pipelines DSLs

almope:
DSLs for the following jobs in the PROD folder:
- rhba-properties-generator
- offline-repo-builder
- rhba-staging

### DIFF
--- a/job-dsls/jobs/prod_offline_repo_builder.groovy
+++ b/job-dsls/jobs/prod_offline_repo_builder.groovy
@@ -7,7 +7,7 @@ String commands = this.getClass().getResource("job-scripts/prod_offline_repo_bui
 
 folder("PROD")
 
-job("offline-repo-builder") {
+job("PROD/offline-repo-builder") {
     description("Check offliner manifest against offline repository")
 
     logRotator {
--- a/job-dsls/jobs/prod_offline_repo_builder.groovy
+++ b/job-dsls/jobs/prod_offline_repo_builder.groovy
@@ -5,6 +5,8 @@ import org.kie.jenkins.jobdsl.Constants
 
 String commands = this.getClass().getResource("job-scripts/prod_offline_repo_builder.sh").text
 
+folder("PROD")
+
 job("offline-repo-builder") {
     description("Check offliner manifest against offline repository")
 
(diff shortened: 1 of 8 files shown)


mbiarnes (approved):
LGTM

job-dsls/jobs/prod_offline_repo_builder.groovy
```diff
@@ -0,0 +1,46 @@
...
+    description("Check offliner manifest against offline repository")
+
+    logRotator {
+        numToKeep(5)
+        artifactNumToKeep(1)
+    }
+
+    parameters {
+        stringParam("MANIFEST_URL", "", "Offliner manifest URL")
+        stringParam("RELEASE_REPO_URL", "http://bxms-qe.rhev-ci-vms.eng.rdu2.redhat.com:8081/nexus/content/repositories/offline-repo-7.6/", "Scratch repository for the release")
```
mareknovotny, line 18: is that URL of repo correct @winklerm ? i am confused by the number 7.6 only ;)

winklerm: Good point - it is outdated; on the other hand, this value is going to be changed for each release, so this particular URL is just a default that needs to be overwritten when running the job.

Anyway, I guess we could put a recent value there, or - maybe better - something like:

```suggestion
        stringParam("RELEASE_REPO_URL", "http://bxms-qe.rhev-ci-vms.eng.rdu2.redhat.com:8081/nexus/content/repositories/offline-repo-REPLACE-ME-WITH-MAJOR.MINOR/", "Scratch repository for the release")
```

Wdyt?

winklerm (approved):
Looks good to me, just one potential improvement (not a blocker from my POV).

job-dsls/jobs/prod_offline_repo_builder.groovy
almope: I went with the update of the parameters to recent values as many times they serve as both a default value and an example. I think it's simpler to understand than using place holders or anything similar.

job-dsls/jobs/seed_job.groovy
```diff
@@ -77,7 +77,10 @@ job("a-seed-job") {
                     "job-dsls/jobs/**/reduced_drools_release.groovy \n" +
                     "job-dsls/jobs/**/prod_projects_downstream_production.groovy \n" +
                     "job-dsls/jobs/**/srcclr_scan_job.groovy\n" +
-                    "job-dsls/jobs/**/jenkins_shared_libs.groovy")
+                    "job-dsls/jobs/**/jenkins_shared_libs.groovy\n" +
+                    "job-dsls/jobs/**/prod_offline_repo_builder.groovy\n" +
```
almope, line 81: @mareknovotny I've added the jobs to the main seed job. Thanks!

PR #934, opened 2021-02-15
Dsl fixes

mbiarnes:

- added missing seed jobs, so we have one for `KIE, KIE-tools, OSBS and PROD`
- renamed seed jobs so they do NOT have all the same name: a-seed-job , rather KIE-seed-job, PROD-seed-job etc.
We only have to build all the seed jobs and move them into the right directories. When they are stored in the right dir they only have to b executed and they will create all jobs in the rights dirs.
-

<pre>

</pre>


job-dsls/jobs/osbs_seed_job.groovy
```diff
@@ -20,7 +20,7 @@ folder("OSBS")
 
 def folderPath="OSBS"
 
-job("${folderPath}/osbs-seed-job") {
+job("${folderPath}/OSBS-seed-job") {
```
mareknovotny, line 23: please keep lower case naming.

PR #956, opened 2021-04-01
replaced origin-repositories.jboss.org

mbiarnes:
 
- replaces the `original-repositories.jboss.org/` URL by `repositories.jboss.org/`
The problem was, that somehow the `repositories.jboss.org` only get updated if there was before a request for certain artifacs.
i.e.: when the artifacts are uploaded to a staging repository for sanity checks these are not visible in `repositories.jboss.org` until there is a request for an artifact. Before i.e. https://repository.jboss.org/nexus/content/groups/kie-group/org/kie/business-central/7.52.0.Final is loaded and accessible the has to be before a request for i.e. https://repository.jboss.org/nexus/content/groups/kie-group/org/kie/business-central/7.52.0.Final/business-central-7.52.0.Final-wildfly-deployable.zip.
Once an artifact is requested the directory gets loaded.
Thus we used before `original-repositories.jboss.org` because there we didn't had this problem, but we were advised to please not use `original-repositories` any more.

- pre-loads some directories in repository.jboss.org so people can access for sanity checks and when they want do download artifacts after the release is closed

<pre>

</pre>

### DIFF
--- a/job-dsls/jobs/kie/master/communityRelease_pipeline.groovy
+++ b/job-dsls/jobs/kie/master/communityRelease_pipeline.groovy
@@ -1,8 +1,8 @@
 import org.kie.jenkins.jobdsl.Constants
 
-def kieVersion=Constants.KIE_PREFIX
+def kieVersion=Constants.KIE_PREFIX + ".Final"
 def baseBranch=Constants.BRANCH
-def releaseBranch="r7.51.0.Final"
+def releaseBranch="r7.53.0.Final"
 def organization=Constants.GITHUB_ORG_UNIT
 def m2Dir = Constants.LOCAL_MVN_REP
 def MAVEN_OPTS="-Xms1g -Xmx3g"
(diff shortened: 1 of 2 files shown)


mbiarnes:
Please don't merge yet - still working on this

job-dsls/jobs/kie/master/communityRelease_pipeline.groovy
```diff
@@ -336,10 +370,10 @@ pipeline {
                 ' \\n' +
                 'You can download the artefacts..: \\n' +
```
winklerm, line 371: A minor typo:
```suggestion
                'You can download the artifacts..: \\n' +
```

winklerm (changes requested):
Please fix the typo, otherwise ok!

From my POV, it would be nice to have the list of URLs defined at one place so future changes will be less error-prone. For instance, after the `wildfly19` suffix changes with a new version. But it is not a blocker for merging this PR.

mareknovotny (approved):

PR #990, opened 2021-05-14
Fix trigger string to accept more loosely based requests.

AntStephenson:
Fix the trigger string to more accepting, for example "jenkins retest please" as seen in: 
https://github.com/kiegroup/kogito-examples/pull/685

### DIFF
--- a/job-dsls/jobs/kie/master/errai_pr.groovy
+++ b/job-dsls/jobs/kie/master/errai_pr.groovy
@@ -82,7 +82,7 @@ job(jobName) {
         githubPullRequest {
             useGitHubHooks(true)
             onlyTriggerPhrase(false)
-            triggerPhrase(".*[j|J]enkins,?.*(retest|test) this.*")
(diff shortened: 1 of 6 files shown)


mbiarnes (approved):
LGTM

mareknovotny:
thanks @AntStephenson we probably need to update also PR template now in https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/.github/pull_request_template.md

AntStephenson:
@mareknovotny What would you prefer to have there? I intentionally didn't update it since "Jenkins retest this" should probably still be the go-to, this just covers those edge cases I guess. if you have suggestions I can make the change.

mareknovotny:
I just would update it as 'this' is not required anymore, but it is true it is not mandatory as it would work even with this

AntStephenson:
Jenkins retest this

mareknovotny (approved):

mbiarnes:
Jenkins retest this

mbiarnes:
Jenkins retest this

PR #1006, opened 2021-06-15
Symlink latest nightly properties

almope:
Add a symlink to the latest nightly properties so it can be referenced without specifying a date


AntStephenson (approved):

Ginxo (approved):
LGTM

PR #1044, opened 2021-07-26
BXMSPROD-1259

almope:

**JIRA**: https://issues.redhat.com/browse/BXMSPROD-1259

After too much time and many considerations I decided to go back to basics and approach the task with this change. The primary objective with the change is to keep it self contained in a single repository and make it easy to follow and understand what the pipeline is really doing. Once a job is created is pretty simple to understand what's happening by just inspecting the pipeline steps. Jenkins shared libraries have to be planned really carefully as they get cluttered pretty easily.

Note: this changes relies on the nightly build and since the rename to `main` branch it hasn't been produced a new nightly so there's not available binaries to run the PR checks and the jobs will fail. This will be fixed on the next nightly successful run.

### DIFF
--- a/job-dsls/jobs/prod/prod_pr_jobs.groovy
+++ b/job-dsls/jobs/prod/prod_pr_jobs.groovy
@@ -1,0 +1,122 @@
+def configs = [:]
+
+configs['izpack'] = [
+        mavenParams: '',
+        pmeParams: '''
+         -DversionOverride=${IZPACK_VERSION}
+        '''
+]
+
+configs['installer-commons'] = [
+        mavenParams: '',
+        pmeParams: '''
+         -DversionOverride=${INSTALLER_COMMONS_VERSION}
+         -DizpackVersion=${IZPACK_VERSION}
+         -DgroovyScripts=file:///${CONFIG_CHECKOUT_PATH}/rhba/nightly/InstallerCommonsAlignment.groovy 
+        '''
+]
+
+configs['rhba-installers'] = [
+        mavenParams: '',
+        pmeParams: '''
+         -DversionOverride=${RHPAM_VERSION}
+         -DizpackVersion=${IZPACK_VERSION}
+         -DinstallerCommonsVersion=${INSTALLER_COMMONS_VERSION}
+         -DkieVersion=${KIE_VERSION}
+         -DgroovyScripts=file:///${CONFIG_CHECKOUT_PATH}/rhba/nightly/InstallerAlignment.groovy,file:///${CONFIG_CHECKOUT_PATH}/rhba/nightly/InstallerCommonsAlignment.groovy,file:///${CONFIG_CHECKOUT_PATH}/rhba/nightly/ForceComRedhatBaVersion.groovy,file:///${CONFIG_CHECKOUT_PATH}/rhba/nightly/ForceComRedhatBaVersionLast.groovy
+        '''
+]
+
+configs['bxms-patch-tools'] = [
+        mavenParams: '-Prhpam,rhdm,integration-tests',
+        pmeParams: '''
+         -DversionOverride=${RHPAM_VERSION} 
+         -DkieVersion=${KIE_VERSION} 
+         -DupdatableVersionRange=7.9.0.redhat-00002,7.9.1.redhat-00003,7.10.0.redhat-00004,7.10.1.redhat-00001
+         -DgroovyScripts=file:///${CONFIG_CHECKOUT_PATH}/rhba/nightly/BxmsPatchToolsPropertiesAlignment.groovy 
+        '''
+]
+
+def folderPath = "PROD/main/pullrequest"
+folder("PROD")
+folder("PROD/main")
+folder("PROD/main/pullrequest")
+
+def commands = this.getClass().getResource("job-scripts/prod_pr_jobs.jenkinsfile").text
+
+configs.each { repository, config ->
+
+    def branchName = repository == 'izpack' ? 'bxms-7.0' : 'main'
+    def parsedCommands = commands.replaceAll(/\$\{(\w+)\}/) { config.containsKey(it[1]) ? config[it[1]] : it[0] }
+
+    pipelineJob("${folderPath}/${repository}-${branchName}.pr") {
+
+        description("Created automatically by Jenkins job DSL plugin. Do not edit manually! The changes will be lost next time the job is generated.\n" +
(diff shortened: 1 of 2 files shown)


almope:
@AntStephenson @Ginxo @mbiarnes @akoufoudakis please, have a look to this when you have a moment. Thanks!

job-dsls/jobs/prod/prod_pr_jobs.groovy
```diff
@@ -0,0 +1,122 @@
...
+def folderPath = "PROD/main/pullrequest"
+folder("PROD")
+folder("PROD/main")
+folder("PROD/main/pullrequest")
+
+def commands = this.getClass().getResource("job-scripts/prod_pr_jobs.jenkinsfile").text
+
+configs.each { repository, config ->
+
+    def branchName = repository == 'izpack' ? 'bxms-7.0' : 'main'
```
Ginxo, line 49: @almope why don't you move branch to `configs`? :thinking:

job-dsls/src/main/resources/job-scripts/prod_pr_jobs.jenkinsfile
```diff
@@ -0,0 +1,101 @@
...
+        stage('Prepare workspace') {
+            steps {
+                cleanWs()
+            }
+        }
+
+        stage('Get version and date for latest successful nightly') {
+            steps {
+                script {
+                    def latestNightly = sh(returnStdout: true, script: "curl -s ${NEXUS_NIGHTLY_REPO}/org/kie/rhba/rhba-parent/maven-metadata.xml | grep '<latest>'")
```
Ginxo, line 32: why not to use same approach as we do here https://github.com/kiegroup/kogito-pipelines/blob/69eed80416aaf1cddee79cfd1e8d57fbf32fda5d/Jenkinsfile.prod.nightly#L44 ? :thinking:

job-dsls/src/main/resources/job-scripts/prod_pr_jobs.jenkinsfile
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+def pmeParameters(bindings) {
+    def params = '''${pmeParams}'''
+    return parseParameters(params, bindings)
+}
+
+def mavenParameters() {
+    def params = '''${mavenParams}'''
```
Ginxo, line 91: sorry but I don't get where `mavenParams` come from... :thinking: can you please explain @almope ?

job-dsls/src/main/resources/job-scripts/prod_pr_jobs.jenkinsfile
```diff
@@ -0,0 +1,101 @@
...
+                    }
+                }
+            }
+        }
+
+    }
+}
+
+def pmeParameters(bindings) {
+    def params = '''${pmeParams}'''
```
Ginxo, line 86: sorry but I don't get where `pmeParams` come from... :thinking: can you please explain @almope ?

Ginxo (changes requested):
please check. Good job!

job-dsls/jobs/prod/prod_pr_jobs.groovy
almope: @Ginxo I wanted to keep the `configs` for the build parameters, both for PME or Maven in this case. But could move the `branch` there if you think it makes more sense to have it there.

job-dsls/src/main/resources/job-scripts/prod_pr_jobs.jenkinsfile
almope: My idea with this file is to serve as a template, but it's true that maybe using the same format for variables and placeholders could be confusing. This, like `mavenParams`, is a placeholder replaced in the `parsedCommands` in here:
https://github.com/kiegroup/kie-jenkins-scripts/pull/1044/files#diff-1ce999c648089b47141457cbdddb69123a6e073e6d06bf531e93fee85551f692R50

job-dsls/src/main/resources/job-scripts/prod_pr_jobs.jenkinsfile
almope: Explained above

job-dsls/jobs/prod/prod_pr_jobs.groovy
Ginxo: I think it's clearer to have it there in the config yes.

Ginxo (approved):
LGTM

job-dsls/jobs/prod/prod_pr_jobs.groovy
almope: Moved to the configs.

job-dsls/src/main/resources/job-scripts/prod_pr_jobs.jenkinsfile
almope: Replaced the placeholders format to `<%= variable %>` to differentiate from the actual pipeline variables.

almope:
@Ginxo added some changes after your suggestions. The change looks even simpler after that. Thanks so much for taking a look! :slightly_smiling_face: Let me know if you think other things have to be revisited.

job-dsls/jobs/prod/prod_pr_jobs.groovy
```diff
@@ -0,0 +1,121 @@
...
+        '''
+]
+
+configs['bxms-patch-tools'] = [
+        branch: 'bxms-7.0',
+        mavenParams: '-Prhpam,rhdm,integration-tests',
+        pmeParams: '''
+         -DversionOverride=${RHPAM_VERSION} 
+         -DkieVersion=${KIE_VERSION} 
+         -DupdatableVersionRange=7.9.0.redhat-00002,7.9.1.redhat-00003,7.10.0.redhat-00004,7.10.1.redhat-00001
```
mareknovotny, line 33: how do we update these ranges? Isn't the maintenance a burner?

almope: It is. The problem is that we need them there for PME to run correctly. They are needed also to run the integration tests. Having them here outdated shouldn't be a problem as the Update Tool would just use them to create a patch and check whether it can be applied using the integration-tests. I have created the below anyway to address that ASAP, so I would merge this in now and later change that.

https://issues.redhat.com/browse/BXMSPROD-1427

mareknovotny (approved):

PR #1047, opened 2021-07-28
changed kie-ci pwd to kie-ci-user-key

mbiarnes:

Added two new parameters to **Constants.groovy** and changed the `kie-ci` PWD by `kie-ci-user-key` SSH key.
Thus, we don't use any more `kie-ci` PWD in kie-jenkins-scripts.

<pre>

</pre>


radtriste:
Beware that using ssh key as creds will need you to setup the url as ssh one instead of https

winklerm:
@mbiarnes For some reason, I cannot submit the review. Anyway, +1 to @radtriste - using the ssh key (kie-ci-user-key) requires to use clone URL with git protocol (not https).

winklerm (changes requested):
Agreed with @radtriste - using kie-ci-user-key requires to use git clone URL with git protocol (ssh). Alternatively, you should be able to use just the token instead of the password.

PR #1062, opened 2021-08-06
Replace undefined variables in PROD PR checks template with an empty string

almope:

When a variable is defined in the PROD PR check jobs template but it's not provided in the configs for the job, it is replaced by a `null` string literal. This is to replace them with an empty string instead.


mareknovotny (approved):
lgtm

PR #1065, opened 2021-08-11
Exclude PNC temporary builds for using them with Prod PR check jobs

almope:


winklerm (approved):

almope:
> so this is more like workaround to status these builds show up in repository, right? the best would be to report it and remove them, Am i right?

@mareknovotny yep, I agree. If that's not necessary for us, or even for Fuse, it shouldn't be there.
@winklerm wdyt?

mareknovotny:
so for now we can merge and later properly fix this. Can you @almope add a comment into code that what you added is a workaround to filter temp build artifacts?

mbiarnes (approved):
LGTM

PR #1066, opened 2021-08-16
Improve ansible roles

Ginxo:
I've noticed several things to improve after RH294 training. Please check. verdaccio and smee related roles are more aligned with ansible-galaxy roles format.

### DIFF
--- a/jenkins-slaves/ansible/data/config/verdaccio/config.yaml.j2
+++ b/jenkins-slaves/ansible/data/config/verdaccio/config.yaml.j2
@@ -27,7 +27,7 @@ web:
 
 auth:
   htpasswd:
-    file: /verdaccio/htpasswd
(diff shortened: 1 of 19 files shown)


mareknovotny (approved):
good you applied new things so soon :)

winklerm (approved):
Very nice, thanks! Good opportunity to learn from your learning  as well :)

Ginxo:
@winklerm @mareknovotny I'm glad you appreciate this change.
In case you are interested, you are able to create a new role just by executing `ansible-galaxy init newrole'' (where `newrole` is the name of the role) and a new folder/files structure is automatically generated.

image

we were using just tasks/main.yml for every role, but it's good we define `vars/main.yml` (with default values) in case we want to share/reuse roles

PR #1102, opened 2021-09-09
BXMSPROD-1443: DSL groovy script for nightlies

mbiarnes:
**Thank you for submitting this pull request**

**JIRA**: _(please edit the JIRA link if it exists)_ 

[BXMSPROD-1443](https://issues.redhat.com/browse/BXMSPROD-1443)

<pre>
How to retest a PR or trigger a specific build:

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>
</pre>

PR #1106, opened 2021-09-16
kogito-tooling job enabled

Ginxo:
according to comment https://github.com/kiegroup/kie-jenkins-scripts/pull/1102#discussion_r705115276 this job is not needed, but I've created PR just in case we want the job to be updated according to https://github.com/kiegroup/kogito-tooling/pull/624

<pre>

</pre>


mareknovotny:
i got the problem here, but maybe we should keep it commented out? do we need it to enable again?

Ginxo:
> i got the problem here, but maybe we should keep it commented out? do we need it to enable again?

no we don't, but at least I have commented out with proper values. ok? thanks @mareknovotny

mareknovotny (approved):

PR #1169, opened 2021-11-19
Enable compile downstream build everywhere

baldimir:
This should enable automatic start of **compile** downstream build (not full downstream built) for all 7.x repositories. I will leave the decision to enable this to others, however I think it should be beneficial.

### DIFF
--- a/job-dsls/jobs/kie/main/compile_downstream_build.groovy
+++ b/job-dsls/jobs/kie/main/compile_downstream_build.groovy
@@ -128,12 +128,7 @@ for (repoConfig in REPO_CONFIGS) {
             pipelineTriggers {
                 triggers {
                     ghprbTrigger {
-                        // execute CDB for drools, droolsjbpm-knowledge and appformer by default
-                        if ((repo != "drools") && (repo != "droolsjbpm-knowledge") && (repo != "appformer")) {
-                            onlyTriggerPhrase(true)
-                        } else {
-                            onlyTriggerPhrase(false)
-                        }
+                        onlyTriggerPhrase(false)
                         gitHubAuthId("${ghAuthTokenId}")
                         adminlist("")
                         orgslist("${ghOrgUnit}")
--- a/job-dsls/jobs/kie/main/compile_downstream_build_7_x.groovy
+++ b/job-dsls/jobs/kie/main/compile_downstream_build_7_x.groovy
@@ -118,11 +118,7 @@ for (repoConfig in REPO_CONFIGS) {
                 triggers {
                     ghprbTrigger {
                         // execute CDB for drools, droolsjbpm-knowledge and appformer by default
-                        if ((repo != "drools") && (repo != "droolsjbpm-knowledge") && (repo != "appformer")) {
-                            onlyTriggerPhrase(true)
-                        } else {
-                            onlyTriggerPhrase(false)
-                        }
+                        onlyTriggerPhrase(false)
                         gitHubAuthId("${ghAuthTokenId}")
                         adminlist("")
                         orgslist("${ghOrgUnit}")


mareknovotny:
i agree it can help, but is this really useful for all repositories - meaning even ones they are not used in any downstream repositories?
I would rather instead of this global one change to add more repos which APIs are used in downstream. Like jbpm, optaplanner or so. We really don't need to enable CDB on all 20+ repositories IMHO.

baldimir:
I agree Marek, that it doesn't make sense for repositories without any dependent repository. I will update the PR.

Ginxo:
Thanks @baldimir for your proposal!!
I agree with @mareknovotny and we would also overload an already overloaded jenkins machine, I'm always about having as much PR checks as possible to help developers to take decisions over a particular change proposal but we have to balance how useful those PR checks really are vs the resources we have for running them.

job-dsls/jobs/kie/main/compile_downstream_build.groovy
```diff
@@ -128,12 +128,7 @@ for (repoConfig in REPO_CONFIGS) {
             pipelineTriggers {
                 triggers {
                     ghprbTrigger {
-                        // execute CDB for drools, droolsjbpm-knowledge and appformer by default
-                        if ((repo != "drools") && (repo != "droolsjbpm-knowledge") && (repo != "appformer")) {
-                            onlyTriggerPhrase(true)
-                        } else {
-                            onlyTriggerPhrase(false)
-                        }
+                        onlyTriggerPhrase(false)
```
mareknovotny, line 131:
```suggestion
                       // execute CDB for all repositories  by default unless it is a leaf repository
                      def repository_leafs = ["lienzo-tests", "kie-wb-distributions", "optaplanner-wb", "jbpm-wb", "openshift-drools-hacep", "process-migration-service", "optaweb-employee-rostering", "optaweb-vehicle-routing", "kie-docs"]
                      if (repository_leafs.contains(repo)) {
                          onlyTriggerPhrase(false)
                      } else {
                          onlyTriggerPhrase(true)
                      }
```

job-dsls/jobs/kie/main/compile_downstream_build_7_x.groovy
```diff
@@ -118,11 +118,7 @@ for (repoConfig in REPO_CONFIGS) {
                 triggers {
                     ghprbTrigger {
                         // execute CDB for drools, droolsjbpm-knowledge and appformer by default
-                        if ((repo != "drools") && (repo != "droolsjbpm-knowledge") && (repo != "appformer")) {
-                            onlyTriggerPhrase(true)
-                        } else {
-                            onlyTriggerPhrase(false)
-                        }
+                        onlyTriggerPhrase(false)
```
mareknovotny, line 121:
```suggestion
                       // execute CDB for all repositories  by default unless it is a leaf repository
                      def repository_leafs = ["lienzo-tests", "kie-wb-distributions", "optaplanner-wb", "jbpm-wb", "openshift-drools-hacep", "process-migration-service", "optaweb-employee-rostering", "optaweb-vehicle-routing", "kie-docs"]
                      if (repository_leafs.contains(repo)) {
                          onlyTriggerPhrase(false)
                      } else {
                          onlyTriggerPhrase(true)
                      }
```

job-dsls/jobs/kie/main/compile_downstream_build.groovy
mareknovotny: i added possible logic for not including leafs repo. But this is just hand coded without checked syntax, so can be wrong ;)

job-dsls/jobs/kie/main/compile_downstream_build_7_x.groovy
mareknovotny: i added possible logic for not including leafs repo. But this is just hand coded without checked syntax, so can be wrong ;)

mbiarnes (approved):
LGTM

job-dsls/jobs/kie/main/compile_downstream_build.groovy
mareknovotny: could we try this suggestion? @baldimir do you agree?

job-dsls/jobs/kie/main/compile_downstream_build_7_x.groovy
mareknovotny: could we try this suggestion? @baldimir do you agree?

mareknovotny:
closing as out dated

PR #1180, opened 2021-12-15
nightly related jobs to priority

Ginxo:
I propose to use a more appropriated label for nightly related jobs in order to speed up nightly jobs. Once the nightly build itself is over the rest of the pipeline is about to trigger the rest of the jobs (send UMB message, generate properties and so on) it takes around 40 minutes and most of that time is about to wait for an available slave instance. What if we open the label filtering to have more possibilities to get a free instance?

 `prod_send_umb_message.groovy` job was missing -> created

Needed PRs:

- https://gitlab.cee.redhat.com/business-automation/eng-jenkins/-/merge_requests/135

<pre>

</pre>

### DIFF
--- a/job-dsls/jobs/prod/prod_cron_meta_nightly_pipeline.groovy
+++ b/job-dsls/jobs/prod/prod_cron_meta_nightly_pipeline.groovy
@@ -6,7 +6,7 @@
 import org.kie.jenkins.jobdsl.Constants
 
 def javadk=Constants.JDK_TOOL
-def AGENT_LABEL="kie-rhel7 && kie-mem4g"
+def AGENT_LABEL="kie-rhel && kie-mem512m"
 
 def NEXT_PRODUCT_VERSION=Constants.NEXT_PROD_VERSION
 def NEXT_PRODUCT_BRANCH='main'
--- a/job-dsls/jobs/prod/prod_cron_meta_nightly_pipeline.groovy
+++ b/job-dsls/jobs/prod/prod_cron_meta_nightly_pipeline.groovy
@@ -125,8 +125,8 @@ pipeline{
 }
 """
 // creates folder if is not existing
-folder("PROD")
 def folderPath="PROD"
+folder(folderPath)
 
 pipelineJob("${folderPath}/cron-meta-nightly-pipeline") {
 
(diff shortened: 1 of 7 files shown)


job-dsls/jobs/prod/prod_kogito_properties_generator.groovy
```diff
@@ -5,7 +5,7 @@
 def kogitoProps ='''
 import groovy.json.JsonOutput
 
-node('kie-rhel7&&!master') {
```
winklerm, line 8: I would suggest to extract label expressions such as this one into a constant(s) so we can change it easily when we need to switch between the clouds (or in case the Jenkins plugin does not work as expected).

winklerm (changes requested):
[Copy from the related eng-jenkins PR comment] Thank you for bringing this up! I am not sure how OpenStack Cloud plugin handles labels resolving to multiple clouds, we had a chat about this use-case with Oliver and my understanding was that there are cases when the plugin will not try the other cloud as expected (it will try the first one and then not try the other one, if the first one fails).

But we can certainly give it a try - I would just propose to extract the label expressions into some common constant(s) so they are changed easily when we need to switch to just one of the clouds.

job-dsls/jobs/prod/prod_cron_meta_nightly_pipeline.groovy
```diff
@@ -6,7 +6,7 @@
 import org.kie.jenkins.jobdsl.Constants
 
 def javadk=Constants.JDK_TOOL
-def AGENT_LABEL="kie-rhel7 && kie-mem4g"
+def AGENT_LABEL="kie-rhel && kie-mem512m"
```
Ginxo, line 9: I don't think we should move this to a constant since it is only used here (very special case)

job-dsls/jobs/prod/prod_kogito_properties_generator.groovy
Ginxo: done, good proposal! thanks

mbiarnes (approved):
LGTM

job-dsls/jobs/prod/prod_cron_meta_nightly_pipeline.groovy
winklerm: Yes, I meant mainly the other cases.

winklerm (approved):
Thanks!

PR #1495, opened 2023-05-02
[BXMSPROD-2016] Migrating dsl-tests from kogito-pipelines to kie-ci repo

bacciotti:


bacciotti:
@lampajr thanks for the advice, really I was in doubt about these files

radtriste:
@lampajr @bacciotti Honestly, I would move the `dsl-tests` action once we move the whole `/dsl` folder to `kie-ci`, not before
Makes more sense to keep both in the same repository

lampajr:
> @lampajr @bacciotti Honestly, I would move the `dsl-tests` action once we move the whole `/dsl` folder to `kie-ci`, not before Makes more sense to keep both in the same repository

Well yes but in that way we would end up having some composite actions here and some others (actually just dsl-tests one) still on kogito-pipelines, I mean I agree that having all dsl stuff in one single place is better but in this case this would be just a reusable composite action that theoretically could be used by other projects as well. Anyway I do not have strong opinion on this so I think it is just up to us :)

radtriste:
ok to move it.
Then we should also soon DSL to `kie-ci`. But first I need to finish https://github.com/kiegroup/kogito-pipelines/pull/775 ..

radtriste:
@bacciotti `jenkins/dsl` file can be removed from this PR, they should not be moved ;)

bacciotti:
@radtriste jenkins/dsl files removed.

radtriste (approved):

lampajr (approved):

PR #1507, opened 2023-05-24
New action for testing

bacciotti:
Not sure if it is the best approach. Just thinking out loud and testing.


bacciotti:
Closing this PR as we had found a better option to solve the issue.

PR #1553, opened 2024-04-29
Weekly triggering of nightlies?

mareknovotny:
Every Wednesday

Changing the triggering of prod nightlies to Wednesdays ONLY?


domhanak (approved):
+1

rgdoliveira:
This will disable all nightlies, not just RHPAM. As we discussed in chat probably the best approach is to split this pipeline in 2 - one for legacy product and other for not legacy products - and then we can properly adjust the cron for each of them.

mareknovotny:
i cloned the `job-dsls/jobs/prod/prod_cron_meta_nightly_pipeline.groovy` into rhpam/bamoe one fyi @rgdoliveira @domhanak

domhanak (approved):

mareknovotny:
ok, so i will merge it now. Is it ok for you @rgdoliveira ?

PR #1572, opened 2024-12-05
print images registries as part of OSL prod UMB message job

rgdoliveira:
Print the images registries in the same format that is expected as part of the promotion email to make it easier.


rgdoliveira:
@mareknovotny just FYI

